PR #58, opened 2026-06-07
Qwen3.6-35B-A3B Q8 GGUF on 2x V100 (+ native MTP spec-decode draft loader)

humanjesse:
Brings up **bartowski's Qwen3.6-35B-A3B Q8_0 GGUF** (35B / 3B active, 256-expert MoE, hybrid Gated-DeltaNet + full-attention, interleaved M-RoPE) as a verified model on 2x V100, and adds a loader for its native **MTP (nextn) speculative-decode head**.

## Main model (no-spec) — Working, verified
Serves coherent, correct output on 2x V100 TP=2 (matches llama.cpp). ~100 tok/s single-stream, ~1900 tok/s aggregate across a 4xTP2 fleet. Three GGUF-interpretation fixes (each affects any GGUF user of arch `qwen35moe`):
- **Gemma RMSNorm double-`+1`** — llama.cpp bakes the `+1` into GGUF norm weights; vLLM re-added it. Subtract 1 at load (excluding gated `linear_attn.norm`).
- **GDN `A_log` double-exponentiation** — GGUF `ssm_a` already stores the decay `A=-exp(A_log)`; store `log(-ssm_a)`.
- **GDN value-head TILE vs repeat_interleave order** — permute value heads pre-shard (byte-clean on Q8_0 boundaries, TP-safe).

## MTP speculative-decode draft — loads, but net-negative on V100
Loads the nextn head from GGUF block 40 as an `mtp` draft (~60% acceptance, exact output within a backend). **Left disabled on V100**: `flash_attn_v100` can't keep CUDA graphs under spec-decode (-> ~46 tok/s); `triton_attn` keeps them but still ~77 vs ~100 no-spec once the draft + 2-token verify overhead is counted. Two loader fixes are general fork bugs affecting **any** GGUF spec-decode draft (honoring the draft's `model_config` in `initialize_model`; stable-sorting the GGUF two-pass weight stream before `AutoWeightsLoader`'s consecutive-only `groupby`). The path would pay off on Ampere/Hopper.

## Regression
No-spec deployed config (FlashAttnV100) re-verified producing correct output with all fixes present. Only the one-line `initialize_model(model_config=...)` touches a shared GGUF path and is a no-op for normal loads (matches `base_loader`); everything else is gated to MTP-only paths.


humanjesse:
Opened against the wrong repo by mistake — this belongs on the humanjesse/vllm-v100 fork, not upstream. Closing.

PR #69, opened 2026-06-19
[Example] Add Higgs-Audio-v3 TTS example for Tesla V100 (SM70)

jajmangold:
## What

Adds a runnable real-time TTS example for `bosonai/higgs-audio-v3-tts-4b` on
Tesla V100 (SM70), under `examples/generate/multimodal/higgs_audio_v3/`:

- `tts.py` — the TTS driver.
- `higgs_v100_low_latency.yaml` — Stage-0 CUDA graph (low-latency) deploy
  profile using `FLASH_ATTN_V100` + `FULL_DECODE_ONLY`.
- `README.md` — setup, requirements, and the eager baseline.

## Why

Higgs-Audio-v3 already runs on V100 in eager mode, but the real-time CUDA graph
path needs two things this example documents:

- the SM70 decode CUDA graph kernel **>= `e64d39aa7`** (already in this fork —
  earlier kernels cap the scalar-paged decode workspace at the capture-time
  `seq_len` and produce incorrect audio under the graph);
- the vllm-omni talker capture fix
  ([vllm-project/vllm-omni#4563](https://github.com/vllm-project/vllm-omni/pull/4563)).

With both, Stage-0 reaches RTF ~1.0 (~2.4x faster than eager); the generated
audio transcribes back to the input prompt.

## Test

Verified on a single Tesla V100: `tts.py` with the included config produces
correct, real-time speech.

### DIFF
--- a/examples/generate/multimodal/higgs_audio_v3/README.md
+++ b/examples/generate/multimodal/higgs_audio_v3/README.md
@@ -1,0 +1,35 @@
+# Higgs-Audio-v3 TTS on Tesla V100 (SM70)
+
+Runs [`bosonai/higgs-audio-v3-tts-4b`](https://huggingface.co/bosonai/higgs-audio-v3-tts-4b)
+text-to-speech on a single Tesla V100 with the `FLASH_ATTN_V100` backend and the
+Stage-0 CUDA graph (low-latency) profile, reaching **real-time** generation
+(RTF ~1.0 — about 2.4x faster than the eager profile).
+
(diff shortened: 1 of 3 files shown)


yangzhuxinyzx:
Closing this example PR because it depends on an unpinned external vllm-omni change and has no reproducible current-head validation. Reopen with pinned compatible dependencies, an executable test/smoke, and updated FLASH_ATTN_V100 deployment evidence.

PR #55, opened 2026-06-01
[V100/SM70] Rollup: Volta serving stack — W4A16 + FP8 (e5m2/MTP) + reasoning fixes

rivetphilbot:
Integration branch bundling all our V100/SM70 serving patches so **one branch** builds a
vLLM that runs **both** compressed-tensors **W4A16** and dense **FP8 (+MTP)** Qwen3.5/Deckard
checkpoints on Volta (2× V100-32GB). Each commit is the clean net diff of an existing focused
PR; this is the all-in-one alternative to merging them piecemeal.

### Bundled (one squashed commit each)
- **#45** — SM70 dense WNA16 TurboMind kernel + V100 (CC 7.0) admittance + DeltaNet loading → **W4A16 on Volta**
- **#49** — skip CT `KVCacheMethod` when the checkpoint has no KV scales → **fp8_e5m2 KV on W4A16** CT
- **#54** — allow `fp8_e5m2` KV with FP8 checkpoints → **fp8_e5m2 KV on dense FP8**
- **#47** — keep partial content when the reasoning block is truncated by `max_tokens` (else empty/cut-off responses)
- **#52** — default the Qwen3 reasoning parser to prompt-has-open-think
- **#46** — CRLF→LF for shell scripts + `.gitattributes`

### Validation
- **FP8 path** (e5m2 + #47) runs live on 2× V100-PCIE serving a dense **FP8 + MTP Qwen3.5-40B**
  at 177K context.
- **W4A16 selection** verified: with #45 the `CompressedTensorsWNA16` scheme admits CC 7.0 and
  W4A16 (asym **and** sym) binds `SM70TurboMindLinearKernel` — stock 1.1.0 admits neither.
- All Python compile-checked; only conflict during assembly was a one-line comment in
  `qwen3_next.py`, resolved.

If you'd rather review/merge atomically, the focused PRs **#45 / #47 / #49 / #54** are open;
this rollup is the convenience superset.

---

## ✅ Verified working configuration (FP8 + MTP, 2× V100)

Validated from a **clean install** (2026-06-01) on 2× V100-PCIE-32GB: 1Cat's prebuilt 1.1.0
wheels + the three **pure-Python** FP8 files from this branch overlaid onto the installed
`vllm/` → the dense **40B FP8 + MTP** checkpoint serves at **177K** with `fp8_e5m2` KV and clean
reasoning separation. No source build for the FP8 path.

**Environment**
[...]

PR #88, opened 2026-06-28
[V100/SM70] fp8_e5m2 KV on W4A16 (P7) + fast fp8 prefill gather + serving docs

rivetphilbot:
### Summary

The two remaining changes needed to run the **full W4A16 + `fp8_e5m2` KV + MTP** stack on 2×V100 (SM70) with 1Cat-vLLM 1.2.1. Everything else in our old rollup (#55) is now in-release; these two are what's left. Both are V100/SM70-only paths and no-ops elsewhere. Includes `docs/v100-sm70-serving.md` with the complete validated serving config.

### Patch 1 — P7: `fp8_e5m2` KV on compressed-tensors W4A16

A W4A16 checkpoint ships **no KV scales** (`kv_cache_scheme=None`), but `get_quant_method` still returns `CompressedTensorsKVCacheMethod`, which **overrides** `process_weights_after_loading`. 1.2.1's SM70 unit-scale override (`attention.py`, `_force_unit_fp8_e5m2_kv_scales`) only engages for the *base* method, so it doesn't apply and boot fails:

```
ValueError: fp8_e5m2 kv-cache is not supported with fp8 checkpoints
outside the SM70 Flash-V100 unit-scale override path.
```

Fix — when no KV scheme ships, skip the CT KV method so the layer takes the plain unit-scale `fp8_e5m2` path:

```python
if isinstance(layer, Attention):
    if self.kv_cache_scheme is None:
        return None
    return CompressedTensorsKVCacheMethod(self)
```

Without this, `--kv-cache-dtype fp8_e5m2` is unusable on W4A16/V100 → no fp8 KV → can't fit the full 262144 context (fp16 KV ~halves the pool).

### Patch 2 — fast `fp8`-KV prefill gather

`_extract_contiguous_kv_from_paged_cache` gates the native `paged_kv_to_contiguous` kernel behind `key_cache.dtype != torch.uint8` (it's fp16-typed), so `fp8` KV falls to a per-block Python loop — the dominant chunked-prefill cost on long prompts. A gather is a bitwise copy, so view two `uint8` bytes as one `float16`, run the fast kernel, view back. **Correctness-equivalent** to the fallback (verified element-wise); decode (incl. MTP) unaffected. Pairs with P7: once e5m2 KV works on W4A16, this keeps its prefill fast.

### Docs

`docs/v100-sm70-serving.md`: what 1.2.1 handles out of the box, both patches above, the **MTP packed-head recipe** (incl. the non-obvious `mtp.fc`-must-stay-fp16 gotcha — a quantized `mtp.fc` silently drops MTP accept to 0%), and the complete recommended serving config.

### Serving config

The full validated config (also in `docs/v100-sm70-serving.md`). Every env var and flag below is required on Volta; `<model>` is any Qwen3.6-27B-class W4A16 checkpoint with a packed (W4A16) MTP head whose `mtp.fc` stays fp16:

```
VLLM_SM70_QUANT_BACKEND=turbomind \
NCCL_P2P_DISABLE=1 NCCL_SHM_DISABLE=0 \
VLLM_WORKER_MULTIPROC_METHOD=spawn \
PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True \
python -m vllm.entrypoints.openai.api_server \
  --model <qwen3.6-27b W4A16 + packed MTP head> \
  --trust-remote-code --dtype half --attention-backend FLASH_ATTN_V100 \
  --tensor-parallel-size 2 --gpu-memory-utilization 0.58 \
  --max-model-len 262144 --max-num-seqs 6 --max-num-batched-tokens 8192 \
  --kv-cache-dtype fp8_e5m2 \
  --speculative-config '{"method":"mtp","num_speculative_tokens":2,"attention_backend":"FLASH_ATTN_V100"}' \
  --compilation-config '{"cudagraph_mode":"full_and_piecewise","cudagraph_capture_sizes":[1,2,4,8]}' \
  --enable-prefix-caching --reasoning-parser qwen3 \
  --enable-auto-tool-choice --tool-call-parser qwen3_coder
```

Notes: `num_speculative_tokens=2` is the sweet spot for stock Qwen MTP on V100 (K=4 lowers accept); `full_and_piecewise` cudagraphs are required so the MTP draft all-reduce is amortized on no-NVLink V100; drop `--kv-cache-dtype fp8_e5m2` (and you no longer need either patch) if you don't need the full 262k context — fp16 KV gives ~192k usable.

### Validation (running in prod, 2×V100-PCIE-32GB, 1.2.1 + both patches)

Qwen3.6-27B-Uncensored W4A16-AWQ g128 + packed MTP head, `fp8_e5m2` KV, **262144** context, MTP K=2:
- **MTP accept ~84%**, **~57 tok/s** single-stream decode (≈1.5× vs no-MTP)
- **427k-token KV pool** at `util 0.58` (1.63× concurrency at 262144), coexisting with an ASR+TTS stack on the same 2 GPUs
- correct outputs at e5m2 unit-scale; long-prompt prefill stays on the fast gather path

### DIFF
--- a/docs/v100-sm70-serving.md
+++ b/docs/v100-sm70-serving.md
@@ -1,0 +1,57 @@
+# Serving Qwen3.5/3.6 on V100 (SM70) with 1Cat-vLLM 1.2.1
+
+Field notes for running W4A16 (+ MTP speculative decoding) Qwen3.6-27B-class checkpoints on
+2x Tesla V100-PCIE-32GB (compute capability 7.0) with 1Cat-vLLM 1.2.1. As of 1.2.1 the Volta
+serving stack is in-release; only the fp8-KV prefill gather (this PR) is an add-on.
+
+## What 1.2.1 already handles (no patch needed)
+- **W4A16 on Volta**: `CompressedTensorsWNA16` admits CC 7.0 and binds the SM70 TurboMind kernel.
+- **fp8_e5m2 KV on W4A16**: `CompressedTensorsKVCacheMethod.validate_kv_cache_scheme(None)` returns
+  cleanly, so `--kv-cache-dtype fp8_e5m2` works on a no-KV-scale W4A16 checkpoint.
+- **SM70 rotary**: pure-PyTorch `forward_native` fallback when `_vllm_fa2_C` is unavailable.
+- **Reasoning**: `--reasoning-parser qwen3` populates the reasoning field and keeps partial content
+  when truncated by `max_tokens`.
+
+## This PR: fast fp8-KV prefill gather
+`_extract_contiguous_kv_from_paged_cache` skips the native `paged_kv_to_contiguous` kernel for
+`uint8` (the kernel is fp16-typed), so fp8 KV falls to a per-block Python loop -- the dominant
+chunked-prefill cost on long prompts for fp8-KV models. The patch views the uint8 bytes as fp16
+(a gather is a bitwise copy) to run the fast kernel. Correctness-equivalent to the fallback;
+only prefill speed changes. Decode (incl. MTP) is unaffected.
+
+## MTP (speculative decoding) packed-head recipe
+To enable `--speculative-config '{"method":"mtp","num_speculative_tokens":2}'` on a W4A16 checkpoint:
+
+- The MTP head must be **quantized (packed W4A16)** to match the body. A bf16 head on a W4A16 body
+  crashes SM70 spec-decode (mixed-precision GDN spec forward).
+- **But `mtp.fc` must stay fp16** and be listed in `quantization_config.ignore` (and excluded from any
+  quant `config_group` target). If `mtp.fc` is marked for quantization while stored plain fp16, the
+  loader logs `Parameter fc.weight not found in params_dict, skip loading`, the drafter's fusion
+  projection is left uninitialized, and **MTP accept rate is 0%** -- silently, since the target model
+  still emits coherent text. With `mtp.fc` in `ignore`, accept is ~76-80%.
+- This matches the base Qwen3.6 packed-head layout: 36 mtp tensors, each `mtp.layers.0.*_proj` with
+  `weight_packed/weight_scale/weight_shape/weight_zero_point`; `mtp.fc` + norms fp16.
+
+## Recommended serving config (measured, 2x V100-PCIE-32GB)
+```
+VLLM_SM70_QUANT_BACKEND=turbomind \
+NCCL_P2P_DISABLE=1 NCCL_SHM_DISABLE=0 \
+VLLM_WORKER_MULTIPROC_METHOD=spawn \
+PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True \
+python -m vllm.entrypoints.openai.api_server \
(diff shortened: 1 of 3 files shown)


Prethea-Phoenixia:
Hello, I came here from the [hugging face repository](https://huggingface.co/philbert440/Qwen3.6-27B-Uncensored-Aggressive-W4A16-AWQ) to report that, after applying the patch on a 1Cat-vLLm install using the config given on hugging face, i am able to run the model locally after passing in --disable-custom-all-reduce to vLLM, otherwise it'll run into a error with the following lines.
```
Failed: Cuda error /home/ymzx/桌面/1cat-vllm/vllm/csrc/custom_all_reduce.cuh:766 'invalid argument'
Failed: Cuda error /home/ymzx/桌面/1cat-vllm/vllm/csrc/custom_all_reduce.cuh:766 'invalid argument'
(EngineCore pid=8958) ERROR 06-29 00:30:17 [multiproc_executor.py:284] Worker proc VllmWorker-1 died unexpectedly, shutting down executor.
```
My configuration is the same as yours -- 2 V100 SXM2 32GB, on Ubuntu 24.04 LTS, etc., is this expected (but undocumented) behavior, or does this indicate some other issues?

Thank you for your work here regardless!

philbert440:
> Hello, I came here from the [hugging face repository](https://huggingface.co/philbert440/Qwen3.6-27B-Uncensored-Aggressive-W4A16-AWQ) to report that, after applying the patch on a 1Cat-vLLm install using the config given on hugging face, i am able to run the model locally after passing in --disable-custom-all-reduce to vLLM, otherwise it'll run into a error with the following lines.
> 
> ```
> Failed: Cuda error /home/ymzx/桌面/1cat-vllm/vllm/csrc/custom_all_reduce.cuh:766 'invalid argument'
> Failed: Cuda error /home/ymzx/桌面/1cat-vllm/vllm/csrc/custom_all_reduce.cuh:766 'invalid argument'
> (EngineCore pid=8958) ERROR 06-29 00:30:17 [multiproc_executor.py:284] Worker proc VllmWorker-1 died unexpectedly, shutting down executor.
> ```
> 
> My configuration is the same as yours -- 2 V100 SXM2 32GB, on Ubuntu 24.04 LTS, etc., is this expected (but undocumented) behavior, or does this indicate some other issues?
> 
> Thank you for your work here regardless!

Expected — and thanks for surfacing it, this is a useful data point. The difference is PCIE vs SXM2. Our validation box is 2× V100-PCIE-32GB, which is cross-socket with no GPU↔GPU P2P, so vLLM auto-disables custom all-reduce at init and that path never runs (hence the NCCL_P2P_DISABLE=1 in the config). Your SXM2 cards have NVLink, so P2P is available, custom all-reduce engages, and on Volta it hits that custom_all_reduce.cuh:766 failure — almost certainly the custom-AR/full-cudagraph-capture interaction.

--disable-custom-all-reduce is the correct fix and the perf cost is negligible: at TP=2 over NVLink the NCCL all-reduce fallback is already fast, and at our decode batch sizes all-reduce isn't the bottleneck. Nothing wrong with the model or the patches. I'll add a note to docs/v100-sm70-serving.md for SXM2/NVLink setups. Thanks for testing it!

PR #96, opened 2026-07-03
sm70/V100: fail loudly when torch<2.10 skips _C_stable_libtorch; tolerate UUID-form CUDA_VISIBLE_DEVICES; add sm_70 build guide

Mermiges:
## Summary

**1. `setup.py`: fail loudly when torch < 2.10 would skip `vllm._C_stable_libtorch`** (90b1cb5)
The stable-ABI extension owns every generic op — activations (`silu_and_mul`, `gelu*`, …), `rms_norm`/`fused_add_rms_norm`, `rotary_embedding`, the entire `_C_cache_ops` namespace, and the CUTLASS scaled_mm/MoE helpers — and only builds against torch >= 2.10. Nothing enforced that at build time (the pin lives in `requirements/cuda.txt`, which `--no-build-isolation` workflows bypass). A wheel produced under older torch is maximally misleading: it installs, imports, and registers all fork-specific kernels, then the first model load dies far from the root cause. This commit turns that into a hard, actionable `RuntimeError` at setup time.

**2. `vllm/platforms`: tolerate UUID-form `CUDA_VISIBLE_DEVICES`** (bf5e135)
The CUDA runtime accepts `GPU-<uuid>` / `MIG-<uuid>` entries in `CUDA_VISIBLE_DEVICES` (the reliable way to pin devices on multi-GPU hosts where indices shift across boots), but `device_id_to_physical_device_id()` ran `int()` on the raw entry. The fix passes UUID entries through unchanged (consumers that rebuild visibility env vars for child processes can use them as-is), raises a clear `ValueError` for anything that is neither an index nor a UUID, and teaches `NvmlCudaPlatform` to resolve UUID entries via `nvmlDeviceGetHandleByUUID` (NVML is unaffected by `CUDA_VISIBLE_DEVICES`, so index-based lookup is wrong for them).

**3. `BUILD_SM70.md`: V100 / sm_70 build and serving guide** (0202f83)
Validated build recipe (torch 2.10.0+cu128 — the cu126 wheel index tops out at 2.9.1, and cu130/CUDA-13 torch drops sm_70; `TORCH_CUDA_ARCH_LIST=7.0`; keep the flash-attention-v100 bundle enabled) plus a serving recipe (float16, device pinning, no `--enforce-eager`) with measured single-V100 numbers.

## Reproduction of the defects

**Defect 1 — silent stable-ABI skip.** Building the wheel in an environment with torch 2.9.1+cu128 (via `pip wheel --no-build-isolation`) succeeded and produced a 100+ MB wheel. `import vllm` worked; all 62 fork-specific sm70 kernels (`awq_*_sm70`, `fp8_*_sm70`, `sm70_f16_*`, marlin, machete) registered in `torch.ops._C`. Serving any model then failed at engine init with:

```
AttributeError: '_OpNamespace' '_C' object has no attribute 'silu_and_mul'
```

Dispatcher inspection showed the `_C` namespace had 62 ops instead of 138 and the `_C_cache_ops` namespace was missing entirely — every generic op (`rms_norm`, `rotary_embedding`, `reshape_and_cache_flash`, …) lives in the never-built `vllm._C_stable_libtorch`. Because import succeeds and the error surfaces at model-load time deep in layer construction, this took real effort to trace back to a build-environment torch version.

**Defect 2 — UUID `CUDA_VISIBLE_DEVICES`.** Launching with `CUDA_DEVICE_ORDER=PCI_BUS_ID CUDA_VISIBLE_DEVICES=GPU-<uuid>` crashed during engine config inside the model-registry inspection subprocess:

```
ValueError: invalid literal for int() with base 10: 'GPU-<uuid>'
```

surfacing to the user as an opaque `Model architectures ['Qwen3ForCausalLM'] failed to be inspected` pydantic ValidationError. Root frame: `vllm/platforms/interface.py`, `device_id_to_physical_device_id()`.

## Validation

- Rebuilt the wheel from this branch's parent (4e9fdbc) with torch 2.10.0+cu128, nvcc 12.6, `TORCH_CUDA_ARCH_LIST=7.0` (~13 min wall on 24 cores): `_C` registers **138 ops** (all 62 sm70 kernels intact + the generic set), `_C_cache_ops` registers 13, and the bundled `flash_attn_v100` exposes all **17** Volta attention features (XQA decode, bhmd, paged prefill bhmd/bfla/splitkv, wmma decode, turboquant decode).
- Runtime-validated on a single V100-SXM2-32GB with Qwen/Qwen3-4B-AWQ (`--dtype float16`, 8192 ctx): `FLASH_ATTN_V100` backend engaged; CUDA graph capture (`mode=VLLM_COMPILE`, `cudagraph_mode=FULL_AND_PIECEWISE`) captured cleanly in 14 s / 0.75 GiB; greedy single-stream decode went from **51.6 t/s (eager) to 132.8 t/s (graphs), a 2.56x lift**, prefill ~3.8k t/s, with **bit-identical greedy output** between eager and graphs.
- Fix 1 verified: `setup.py` under torch 2.9.1 now aborts immediately with the actionable message instead of building a broken wheel.
- Fix 2 verified: with `CUDA_VISIBLE_DEVICES=GPU-<uuid>`, `device_id_to_physical_device_id(0)` passes the UUID through, `NvmlCudaPlatform.get_device_capability(0)` resolves `DeviceCapability(major=7, minor=0)` via NVML, and the previously-crashing model-module import path completes.

Measurements come from PLI Labs V100 serving research (proprietarylegal.ai).

### DIFF
--- a/BUILD_SM70.md
+++ b/BUILD_SM70.md
@@ -1,0 +1,75 @@
+# Building and serving 1Cat-vLLM on NVIDIA V100 (sm_70)
+
+Validated recipe for building this fork from source for Volta (V100-32GB) and
+serving with the compiled `FLASH_ATTN_V100` backend. All numbers below were
+measured on a single V100-SXM2-32GB.
+
+## Build environment
+
+| requirement | value | why |
+|---|---|---|
+| Python | 3.12 | tested combination |
+| torch | **2.10.0+cu128** | the fork pins `torch==2.10.0` (`requirements/cuda.txt`); the stable-ABI op library `vllm._C_stable_libtorch` requires torch >= 2.10. The cu126 wheel index tops out at torch 2.9.1, so cu128 is the correct CUDA variant. torch 2.10.0+cu128 still ships sm_70 kernels — verify with `python -c "import torch; print(torch._C._cuda_getArchFlags())"` (must include `sm_70`). Do **not** use a CUDA 13 (cu130) torch: sm_70 is dropped there. |
+| CUDA toolkit (nvcc) | 12.6+ works | minor-version mismatch vs. torch's cu128 runtime is fine for compilation |
+| `TORCH_CUDA_ARCH_LIST` | `7.0` | compile only sm_70; keeps the build to ~13 min on a 24-core host |
+
+Do **not** set `VLLM_SKIP_FLASH_ATTN_V100_BUNDLE` — the bundled
+`flash-attention-v100` extension is what provides the Volta attention kernels
+(XQA decode, bhmd, paged prefill bhmd/bfla/splitkv, wmma decode, turboquant
+decode). A wheel built without the bundle, or installed next to a stale
+`flash_attn_v100` package, makes the strict `FLASH_ATTN_V100` backend refuse
+to run ("required Flash op is unavailable").
+
+```bash
+export TORCH_CUDA_ARCH_LIST=7.0
+export FLASH_ATTN_V100_CUDA_ARCH_LIST=7.0
+export MAX_JOBS=<cores>
+pip wheel --no-build-isolation --no-deps -w dist .
+pip install dist/1cat_vllm-*.whl
+```
+
+If the build environment has torch < 2.10, `setup.py` now fails loudly instead
+of producing a wheel whose `_C` namespace is missing every generic op
+(`silu_and_mul`, `rms_norm`, `rotary_embedding`, all of `_C_cache_ops`, ...) —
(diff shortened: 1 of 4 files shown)


yangzhuxinyzx:
Closing the stale mixed-scope PR. It is based nine main commits back and combines a build-policy change, UUID handling, and documentation without focused tests on current main. The UUID behavior and any stable-libtorch build guard should be reintroduced as separate current-main PRs with unit/build-matrix coverage.

PR #104, opened 2026-07-18
[V100/SM70] Gemma4 complete adaptation: FA head_dim=512 + MTP speculative decoding + vision encoding OOM fix + compile priming

lwh9346:
## Purpose

Complete **gemma-4-31B** adaptation for V100/SM70 with the **FLASH_ATTN_V100** backend:

1. **Kernel**: the hybrid attention model (50× sliding-window layers at head_dim 256 + 10× global layers at **head_dim 512**, unified K/V) aborted at startup with `Unsupported head_dim for paged decode: 512`.
2. **Compile**: a startup crash that the SM70 0.0.3 compile graph policy exposed for any model whose first compiled batch is 1.
3. **MTP speculative decoding**: the Gemma4 MTP proposer crashed on every draft step beyond the first, and the draft metadata build stalled the CPU on a blocking D2H copy every step.
4. **Multimodal**: high-soft-token vision encoding OOMed the worker on SM70.

### 1. Kernel: head_dim=512 instantiation (24 lines across 3 files)

Previously the backend aborted at startup with `Unsupported head_dim for paged decode: 512`.

- `flash-attention-v100/kernel/flash_decode_paged.cu` — dispatch `case 512` for paged decode
- `flash-attention-v100/kernel/fused_mha_forward_paged.cu` — `Config<512>` (BLOCK_M=16, BLOCK_N=32, 16 warps) for paged prefill
- `flash-attention-v100/kernel/fused_mha_forward.cu` — same tile config for dense prefill
- Adds `flash-attention-v100/test_fa512.py` (parity harness, see Test Plan)

### 2. Compile: prime torch.compile above batch 1 (`vllm/v1/worker/gpu_worker.py`)

With the kernel adapted, startup still crashed once the 0.0.3 policy was auto-enabled: the compiled model returned `hidden_states` shaped `[1, H]` for num_tokens>1 warmup dummy runs, tripping a device-side assert in `gpu_model_runner._dummy_run`'s `hidden_states[logit_indices]` gather.

**Root cause (reproduced in a standalone torch experiment):** with BACKED dynamic shapes (the default), torch.compile specializes a symbol whose first-seen value is 0 or 1. CUDA-graph capture compiles largest-first, so configs with a capture size ≥ 2 are safe — but with `cudagraph_capture_sizes=[1]` (what our SM70 serve configs use) the *first* compiled forward runs at num_tokens=1 and batch=1 is baked into the artifact. Since vLLM drops dynamo guards by design (and the 0.0.3 policy's AOT path cannot recompile on a guard miss), every later call reuses the poisoned artifact. The defect is model-agnostic; Gemma4 was simply the first model deployed here with an all-1 compile config.

**Fix:** in `compile_or_warm_up_model`, when all compiled sizes are 1 while the scheduler can batch beyond one token, append a priming warmup at num_tokens=2 so the first trace keeps the batch dimension dynamic. Size-1 capture/replay and larger non-captured runs then both behave. This replaces the interim Gemma4-only policy guard (693a295f6, dropped from this branch); the policy now auto-enables for Gemma4 as intended.

*Residual note:* a config with no startup-time compilation at all (no capture/compile sizes) could still first-compile on a real batch-1 request — that upstream-wide hazard is out of scope here.

### 3. MTP speculative decoding fixes (`vllm/v1/spec_decode/gemma4.py`, `vllm/v1/attention/backends/flash_attn_v100.py`)

**Crash fixes:** `Gemma4Proposer._greedy_sample` did not accept the `spec_step_idx` argument passed by the base proposer loop, crashing every draft step beyond the first. `Gemma4MTP.get_top_tokens` also does not take `spec_step_idx`, so `use_local_argmax_reduction` is disabled to route through the compatible `super()._greedy_sample` path.

**D2H sync stall:** `_attach_common_flash_metadata` fell back to the deprecated `seq_lens_cpu` property whenever the `_seq_lens_cpu` shadow was unset. The spec-decode draft loop invalidates that shadow every step (`llm_base_proposer.py`), so each draft iteration issued a **blocking `seq_lens.to("cpu")` copy** that stalled the CPU behind the in-flight target forward (~12.6 ms per propose call at batch 1). The fix uses the runner-maintained `seq_lens_cpu_upper_bound` instead — kept on CPU, no sync, and sufficient for every consumer on the draft decode path (decode shape hints and prefix-context detection); the decode kernels themselves read exact lengths from device tensors. SM70 MTP proposer profile (batch=1): **loop_metadata_cpu 12.6 ms → 0.24 ms**, proposer wall 17.2 ms → 4.8 ms.

**Deployment note:** `cudagraph_capture_sizes` must be multiples of `num_speculative_tokens + 1`. With MTP and `[1,2,3,4]`, sizes round up to just `{4}` (single-request verify only) — concurrency ≥2 silently falls off the FULL cudagraph path and aggregate throughput collapses. Working config:

```
--compilation-config '{"cudagraph_mode":"full_and_piecewise","cudagraph_capture_sizes":[4,8,12,16]}' \
--speculative-config '{"method":"mtp","model":"<assistant>","num_speculative_tokens":3}'
```

Decode throughput (tok/s, greedy, 512 output tokens, TP=8, W4A16), greedy output **byte-identical** to non-MTP (lossless verified):

| Concurrency | Non-MTP | MTP=3 | Speedup |
|---|---|---|---|
| 1 | 60.1 | 120.8 | **2.0×** |
| 2 | 118.3 | 236.2 | **2.0×** |
| 4 | 232.1 | 417.1 | **1.8×** |

### 4. Multimodal: high-soft-token vision encoding memory-safe on SM70 (`vllm/model_executor/models/gemma4_mm.py`)

Two compounding defects OOMed the worker (crashing the engine) when encoding images at high max_soft_tokens (e.g. 1120, i.e. 10080 patches):

1. The HF vision/audio towers stay in the checkpoint's bf16 dtype — the fp16 cast applied to the vLLM language stack does not cover them. torch's memory-efficient SDPA kernel rejects bf16 on SM70 (fp16/fp32 only), so tower attention always fell back to the math backend and materialized N×N fp16 scores per head: a ~6 GiB single allocation for 2 images at 10080 patches, exceeding the ~1-2 GiB free outside the vLLM memory pool. Cast the towers (and embedders) to the language model's dtype in `load_weights`.
2. The 2D bool padding mask passed to the HF encoder is expanded internally to a full `(B, 1, N, N)` bool mask, which likewise forces the math SDPA backend. Build the additive key-padding mask directly in the query dtype with broadcastable shape `(B, 1, 1, N)`, keeping SDPA on the memory-efficient backend; outputs are bit-identical to the bool-mask path (verified offline), and `None` is passed when a bucket has no padding.

### Relationship to prior work (duplicate-work check)

- Checked open PRs/issues for "head_dim", "Gemma4/gemma-4", "512": only #77 (fp8 KV gather) is open — unrelated.
- **#59** ([V100/SM70] gemma-4 support, fully-FA hybrid attention) covered D=512 inside a much larger model+MTP enablement PR and was **closed unmerged** (2026-06-14). The sliding-window FA support from that line of work has since landed in the tree; the D=512 instantiation never did. This PR now carries the complete Gemma4 adaptation: kernel, compile, MTP, and multimodal.

## Test Plan

- **Kernel parity**: `flash-attention-v100/test_fa512.py` — 7 cases vs fp32 reference: dense prefill / paged prefill / paged prefill-chunked / paged decode at D=512, plus D=256 regression on all three paths. Inputs harvested from real gemma-4 `embed_tokens` rows (`--model-dir`, with seeded-random fallback).
- **Compile fix**: standalone torch repro of the 0/1 specialization (mark_dynamic + size-1 first compile + guard-free replay ⇒ size-1 outputs); then end-to-end: serve `gemma-4-31B-it-qat-w4a16-ct` (compressed-tensors W4A16, SM70 TurboMind path) with FLASH_ATTN_V100, `cudagraph_capture_sizes=[1]`, 0.0.3 policy auto-enabled — startup and chat-template generation at TP4/TP8.
- **MTP**: serve with `--speculative-config '{"method":"mtp","num_speculative_tokens":3}'`; verify no proposer crashes across long generations; greedy output compared byte-for-byte against the non-MTP baseline (identical ⇒ lossless); acceptance rate from `/metrics` (pos0 86%, pos1 74%, pos2 66%, ~3.2 tokens/step).
- **Vision**: `max_soft_tokens=1120` with 1/2/4 images per prompt — all encode and answer correctly where 2 images previously OOMed with a 6 GiB allocation. Combined with `gpu-memory-utilization=0.80` for encoder headroom.

## Test Result

Parity: **7/7 PASS** (max_abs ≤ 0.016 vs fp32 ref; D=256 paths unchanged).

Compile fix: policy auto-enables for Gemma4, priming warmup at size 2 fires, startup clean, greedy chat answers correct (`2+2` → `4`) at TP4; TP8 service verification in the comments below.

TP8, single request, prefill/decode tok/s — FLASH_ATTN_V100 vs TRITON_ATTN baseline (identical serving config):

| ctx | Triton prefill | FA prefill | speedup | Triton decode | FA decode | speedup |
|----:|----:|----:|----:|----:|----:|----:|
| 1024 | 1544.69 | 3209.64 | **2.1×** | 59.67 | 60.55 | 1.01× |
| 2048 | 1575.29 | 4828.49 | **3.1×** | 51.42 | 60.37 | 1.17× |
| 4096 | 995.07 | 4653.80 | **4.7×** | 40.46 | 59.95 | 1.48× |
| 8192 | 593.58 | 3818.40 | **6.4×** | 28.36 | 59.29 | **2.1×** |
| 16384 | 330.37 | 2905.40 | **8.8×** | 21.67 | 57.15 | **2.6×** |
| 32700 | 178.03 | 2120.39 | **11.9×** | 13.60 | 51.80 | **3.8×** |

TP4: FA completes all context lengths (32700 at 1359 t/s prefill / 46.0 t/s decode) where the Triton TP4 baseline timed out (>900 s); FA decode barely degrades with context length (60.6 → 51.8 from 1k → 32k) vs Triton (59.7 → 13.6).

---

**AI-assistance disclosure**: this change was developed and tested with AI assistance (Kimi Code). The submitting human has reviewed the diff and reproduced the key results before marking ready for review.

### DIFF
--- a/flash-attention-v100/flash_attn_v100/__init__.py
+++ b/flash-attention-v100/flash_attn_v100/__init__.py
@@ -3,6 +3,8 @@ __version__ = "1.2.0"
 from flash_attn_v100.flash_attn_interface import (
     flash_attn_decode_qk_scores,
     flash_attn_decode_paged,
+    flash_attn_decode_paged_mq,
+    flash_attn_decode_paged_mq_available,
     flash_attn_decode_paged_xqa,
     flash_attn_decode_paged_xqa_available,
     flash_attn_decode_paged_wmma,
@@ -21,6 +23,8 @@ from flash_attn_v100.flash_attn_interface import (
 __all__ = [
     "flash_attn_decode_qk_scores",
     "flash_attn_decode_paged",
+    "flash_attn_decode_paged_mq",
+    "flash_attn_decode_paged_mq_available",
     "flash_attn_decode_paged_xqa",
     "flash_attn_decode_paged_xqa_available",
     "flash_attn_decode_paged_wmma",
--- a/flash-attention-v100/flash_attn_v100/flash_attn_interface.py
+++ b/flash-attention-v100/flash_attn_v100/flash_attn_interface.py
@@ -15,8 +15,10 @@ except ImportError:
 
 DEFAULT_DECODE_PARTITION_SIZE = 256
 VALID_DECODE_PARTITION_SIZES = (256, 512, 1024)
+DECODE_MQ_Q_MAX = 8
 _decode_plan_cache = {}
 _decode_workspace_cache = {}
+_decode_mq_workspace_cache = {}
 _turboquant_decode_workspace_cache = {}
 logger = logging.getLogger(__name__)
 
@@ -134,6 +136,7 @@ def _get_decode_plan(
     batch_size_hint: Optional[int] = None,
     workspace_seq_capacity_hint: Optional[int] = None,
     active_num_partitions: Optional[torch.Tensor] = None,
+    default_partition_size: Optional[int] = None,
 ) -> _DecodePlan:
     batch_capacity = batch_size_hint or block_table.shape[0]
     num_heads = q.shape[1]
@@ -169,6 +172,7 @@ def _get_decode_plan(
         num_kv_heads=k_cache.shape[2],
         max_seq_len_hint=effective_max_seq_len,
         batch_size_hint=batch_capacity,
+        default_partition_size=default_partition_size,
     )
     runtime_num_partitions = max(
         1,
@@ -256,6 +260,102 @@ def _get_decode_workspace_for_plan(
     )
 
 
+@dataclass
+class _DecodeWorkspaceMq:
+    tmp_out: torch.Tensor
+    max_logits: torch.Tensor
+    exp_sums: torch.Tensor
+    active_num_partitions: torch.Tensor
+    max_num_partitions: int
+
+
+def _allocate_decode_workspace_mq(
+    q: torch.Tensor,
+    *,
+    batch_capacity: int,
+    num_heads: int,
+    head_dim: int,
+    q_max: int,
+    max_num_partitions: int,
+) -> _DecodeWorkspaceMq:
+    return _DecodeWorkspaceMq(
+        tmp_out=torch.empty(
+            (batch_capacity, num_heads, q_max, max_num_partitions, head_dim),
+            dtype=torch.float16,
+            device=q.device,
+        ),
+        max_logits=torch.empty(
+            (batch_capacity, num_heads, q_max, max_num_partitions),
+            dtype=torch.float32,
+            device=q.device,
+        ),
+        exp_sums=torch.empty(
+            (batch_capacity, num_heads, q_max, max_num_partitions),
+            dtype=torch.float32,
+            device=q.device,
+        ),
+        active_num_partitions=torch.empty(
+            (1,),
+            dtype=torch.int32,
+            device=q.device,
+        ),
+        max_num_partitions=max_num_partitions,
+    )
+
+
+def _get_decode_workspace_mq_for_plan(
+    q: torch.Tensor,
+    *,
+    batch_capacity: int,
+    num_heads: int,
+    head_dim: int,
+    plan: _DecodePlan,
+    active_num_partitions: Optional[torch.Tensor] = None,
+):
+    device_index = q.device.index if q.device.index is not None else -1
+    stream_id = _workspace_stream_id(q.device)
+    key = (
+        "mq",
+        device_index,
+        stream_id,
+        batch_capacity,
+        num_heads,
+        head_dim,
+        DECODE_MQ_Q_MAX,
+        plan.partition_size,
+    )
+
+    workspace = (
+        _decode_mq_workspace_cache.get(key) if _can_cache_workspace(q) else None
+    )
+    if (
+        workspace is None
+        or workspace.max_num_partitions < plan.workspace_num_partitions
+    ):
+        workspace = _allocate_decode_workspace_mq(
+            q,
+            batch_capacity=batch_capacity,
+            num_heads=num_heads,
+            head_dim=head_dim,
+            q_max=DECODE_MQ_Q_MAX,
+            max_num_partitions=_round_decode_partition_capacity(
+                plan.workspace_num_partitions
+            ),
+        )
+        if _can_cache_workspace(q):
+            _decode_mq_workspace_cache[key] = workspace
+
+    if active_num_partitions is None:
+        workspace.active_num_partitions.fill_(plan.actual_num_partitions)
+        active_num_partitions = workspace.active_num_partitions
+    return (
+        workspace.tmp_out[:, :, :, :workspace.max_num_partitions, :],
+        workspace.max_logits[:, :, :, :workspace.max_num_partitions],
+        workspace.exp_sums[:, :, :, :workspace.max_num_partitions],
+        active_num_partitions,
+    )
+
+
 def _get_turboquant_decode_workspace(
     q_rot: torch.Tensor,
     kv_cache: torch.Tensor,
@@ -321,9 +421,15 @@ def _get_decode_partition_size(
     num_kv_heads: int,
     max_seq_len_hint: Optional[int] = None,
     batch_size_hint: Optional[int] = None,
+    default_partition_size: Optional[int] = None,
 ) -> int:
     raw = os.getenv("VLLM_FLASH_V100_DECODE_PARTITION_SIZE")
     if raw is None:
+        if default_partition_size is not None:
+            return _validate_decode_partition_size(
+                default_partition_size,
+                "default_partition_size",
+            )
         return _select_default_decode_partition_size(max_seq_len_hint)
     try:
         value = int(raw)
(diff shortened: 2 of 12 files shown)


lwh9346:
## Gemma4 MTP speculative decoding: crash fixes + D2H sync stall fix + benchmark data

Two new commits pushed to this PR:

### 1. `[Bugfix] gemma4 MTP: fix proposer crashes in speculative decode` (4af190d7f)

MTP (`--speculative-config '{"method":"mtp", ...}'`) crashed on every draft step beyond the first:
- `Gemma4Proposer._greedy_sample` missing the `spec_step_idx` argument
- `Gemma4MTP.get_top_tokens` doesn't accept `spec_step_idx` → `use_local_argmax_reduction = False`

### 2. `[Perf] flash-attention-v100: avoid blocking D2H sync in draft metadata build` (6480701c0)

`_attach_common_flash_metadata` fell back to the deprecated `seq_lens_cpu` property when the `_seq_lens_cpu` shadow was unset. The spec-decode draft loop invalidates that shadow every step (`llm_base_proposer.py`), so every draft step issued a **blocking `seq_lens.to("cpu")` D2H copy** that stalled the CPU behind the in-flight target forward.

Fix: use the runner-maintained `seq_lens_cpu_upper_bound` (always on CPU, no sync). Sufficient for every consumer on the draft decode path — decode kernels read exact lengths from device tensors; only shape hints use the CPU copy.

Measured via built-in SM70 MTP proposer profile (batch=1): **loop_metadata_cpu 12.6ms → 0.24ms**, proposer wall 17.2ms → 4.8ms.

### Benchmarks (Gemma4-31B W4A16, TP=8, 8×V100, MTP=3)

Decode throughput (tok/s), greedy, 512 output tokens:

| Concurrency | Non-MTP | MTP=3 (this PR) | Speedup |
|---|---|---|---|
| 1 | 60.1 | 120.8 | **2.0×** |
| 2 | 118.3 | 236.2 | **2.0×** |
| 4 | 232.1 | 417.1 | **1.8×** |

Correctness: greedy output with MTP is **byte-identical** to non-MTP output (lossless verified).

### Important deployment note

`cudagraph_capture_sizes` must be multiples of `num_speculative_tokens + 1`. With MTP and `[1,2,3,4]`, sizes round up to just `{4}` (single-request verify only) — concurrency ≥2 silently falls off the FULL cudagraph path and aggregate throughput collapses (~80 tok/s at conc=2). Working config:

```
--compilation-config '{"cudagraph_mode":"full_and_piecewise","cudagraph_capture_sizes":[4,8,12,16]}' --speculative-config '{"method":"mtp","model":"<assistant>","num_speculative_tokens":3}'
```

(sizes 4/8/12/16 = 1–4 requests × 4 verify tokens)

lwh9346:
## Full benchmark: TP2 / TP4 / TP8 — prefill + decode, context & concurrency scaling

**Setup**: gemma-4-31B-it-qat-w4a16-ct (W4A16), FLASH_ATTN_V100, MTP=3 (except TP2, see notes), greedy, 256 output tokens, exact prompt token counts via the gemma4 tokenizer (unique random header per request to defeat prefix caching). 8× Tesla V100 16GB (DGX-1 topology: GPU0-3 NUMA0/NVLink, GPU4-7 NUMA1/NVLink).

### TP8 (GPUs 0-7, MTP=3, max_model_len 262144, KV cache 376k tokens)

| Context | Concurrency | Prefill tok/s (per req) | Decode tok/s (per req) |
|---|---|---|---|
| 1k | 1 | 3534 | 110.7 |
| 8k | 1 | 3560 | 114.6 |
| 32k | 1 | 1957 | 90.0 |
| 128k | 1 | 819 | 39.5 |
| 8k | 2 | 2013 (agg ~4.0k) | 65.6 (agg ~131) |
| 8k | 4 | 1372 (agg ~5.5k) | 47.7 (agg ~191) |
| 32k | 2 | 1402 (agg ~2.8k) | 42.7 (agg ~85) |
| 32k | 4 | 985 (agg ~3.9k) | 18.2 (agg ~73) |
| 128k | 2 | 619 (agg ~1.2k) | 17.8 (agg ~36) |

Mixed load (1× 32k prefill + 3× 1k decode concurrently): prefill 1977 tok/s, decode 14.0 tok/s/req during the prefill storm (decode recovers to normal after).

### TP4 (GPUs 0-3, MTP=3, max_model_len 131072, KV cache 185k tokens)

| Context | Concurrency | Prefill tok/s | Decode tok/s (per req) |
|---|---|---|---|
| 1k | 1 | 2134 | 110.6 |
| 8k | 1 | 2238 | 111.3 |
| 32k | 1 | 1271 | 67.5 |
| 128k | 1 | 469 | 32.0 |
| 8k | 2 | 1321 | 61.4 |
| 8k | 4 | 887 | 38.5 |
| 32k | 2 | 883 | 28.9 |
| 32k | 4 | 605 | 12.5 |
| 128k | 2 | serialized (2×128k > KV) | 33.9 (one at a time) |

Mixed load: prefill 1184 tok/s, decode 8.6 tok/s/req during prefill.

### TP2 (GPUs 4-5, **no MTP**, max_model_len 8192, gpu-mem-util 0.92, KV cache 8.5k tokens)

| Context | Concurrency | Prefill tok/s | Decode tok/s (per req) |
|---|---|---|---|
| 1k | 1 | 1185 | 40.2 |
| 7.7k | 1 | 1216 | 38.6 |
| 1k | 2 | 673 | 39.1 |
| 1k | 4 | 473 | 33.1 |

### Notes

- **Decode at concurrency 1 is identical on TP4 and TP8** (~111-115 tok/s): single-request decode is dominated by fixed per-step overheads, not weight-read bandwidth. TP2 is much slower (40 tok/s) — 4× weight bytes per GPU and no MTP (MTP does not fit in memory at TP2).
- **Prefill scales sub-linearly but well**: 1185 → 2238 → 3560 tok/s for TP2 → TP4 → TP8 (1.9× / 1.6× per doubling).
- **Long-context decode degrades with KV reads**: TP8 decode 114.6 (8k) → 90.0 (32k) → 39.5 (128k) tok/s.
- **TP2 feasibility**: the full stack (31B W4A16 + multimodal tower + MTP assistant) at 0.92 gpu-mem-util leaves only ~1.9 GiB KV per GPU (~4.5k context with MTP; 8.5k without). TP2 works for short contexts only; MTP was dropped to get a usable KV pool.
- **fp8_e5m2 KV cache**: not usable with this checkpoint — `vllm/model_executor/layers/attention/attention.py:180` rejects fp8_e5m2 KV for compressed-tensors W4A16 models outside the SM70 Flash-V100 unit-scale override path (that path requires an FP8 checkpoint with base scale processing). README's experimental fp8 note therefore does not apply to this model; omitted from the matrix.
- MTP acceptance (TP8, from /metrics): pos0 86%, pos1 74%, pos2 66%, ~3.2 tokens/step average; greedy output byte-identical to non-MTP.

lwh9346:
## Multi-query paged decode for MTP verify (798a09ba0)

**Problem**: the smallq decode path folded verify q_len into the batch dim — KV cache was read q_len times per step (4× at MTP=3, 8× with GQA). At 128k, full-attn verify cost 13×2.52ms=32.8ms/step and decode collapsed to 39.5 tok/s.

**Fix**: new `flash_attn_decode_paged_mq` op — one CTA per (request, head, partition) applies all q_len query tokens of the request against each K/V tile (per-query causal limits + sliding-window bounds). MQ_QLEN {1,2,4,8} instantiations → MTP1–MTP7; head_dim 256 (sliding) and 512 (full-attn); fp16/fp8 KV; CUDA-graph-safe persistent metadata; sliding-window layers stay folded (fixed 1k window → too few CTAs for mq). Fallback: `VLLM_FLASH_V100_SMALLQ_FOLD=1`.

**Correctness**: kernel parity vs folded reference bit-identical at matched partition size (~120 cases: seq 1k–128k, q_len 1–8, batch 1–4, window on/off, padding, interleaved KV); E2E greedy 256-token outputs byte-identical 10/10 prompts.

**Perf (TP8, same-box A/B)**:
| cell | folded | mq | Δ |
|---|---|---|---|
| 128k×1 decode | 39.5–43.9 | 65.5 tok/s | +50~55% |
| 128k×2 decode | 16.2 | 27.1 tok/s/req | +67% |
| full-attn verify call | 2.52ms | 1.42ms | −44% |
| 32k×2 / ×4 | 40.3 / 18.5 | 44.7 / 23.2 | +11% / +25% |
| 1k/8k/32k×1 | — | — | no regression (noise) |

**MTP sweep with mq (≤32k, conc≤2, median)**: 32k×1 82 → 107 (MTP5) → 120 (MTP7) tok/s; 8k×2 80 → 82.5 → 92.3; 1k×1 102 → 97 → 117.6; acceptance length 3.55 / 4.57 / 5.48 at MTP3/5/7. 128k×1: 65.5 (MTP3) → 53.1 (MTP5) → 52.1 (MTP7) — MTP≥5 pays ~20% at 128k. Note: num_speculative_tokens > 3 needs `cudagraph_capture_sizes` in multiples of num_spec+1 (MTP=7 → [8,16,24,32]).

yangzhuxinyzx:
Closing this draft rather than attempting a conflict-heavy merge.\n\nThe branch is conflicting with current main and combines independent changes: Flash-V100 D=512 kernels, compiler warmup policy, Gemma MTP compatibility, draft metadata synchronization, multi-query decode, and multimodal memory behavior. Each needs its own current-head baseline, correctness/quality gate, and SM70 build validation. Please split and rebase the individual changes before reopening.

PR #116, opened 2026-08-01
[Bugfix] Fix SM70 TileLang JIT compile (#105)

### LINKED ISSUE #105: [Bug]: SM70 default TileLang prefill fails to compile with tilelang==0.1.9
### Summary

On V100 / SM70, the default FlashQLA TileLang GDN prefill route in the 1Cat-vLLM 1.2.2 stack deterministically fails at JIT compilation.

The immediate cause is the pinned `tilelang==0.1.9`. Its CUDA header has a broken BF16 fallback for `__CUDA_ARCH__ < 800`: `common.h:778` passes `__nv_bfloat16` values to the half-only intrinsic `__hfma`.

### Environment

- 1Cat-vLLM 1.2.2 wheel / CUDA 12.8 image
- Python 3.12, CUDA toolkit 12.8.93
- 2 x Tesla V100 32 GB (compute capability 7.0), TP=2
- `tilelang==0.1.9`
- Qwen3.6-27B AWQ hybrid GDN model
- `gdn_prefill_backend=flashqla_sm70`
- `VLLM_SM70_FLASHQLA_ORIGINAL_PREFILL` left unset (the current code defaults it to the original TileLang route)

The repository currently pins both `apache-tvm-ffi==0.1.9` and `tilelang==0.1.9` in [requirements/cuda.txt](https://github.com/1CatAI/1Cat-vLLM/blob/644d8a7cd05ed4ecd1cd188e3c05b4bbd074f504/requirements/cuda.txt).

### Actual result

The first GDN prefill warmup / request selects the original SM70 TileLang path and tries to JIT `tilelang_chunk_local_cumsum_kernel` with `nvcc -arch=sm_70`. Compilation then fails at:

```text
.../tilelang/src/tl_templates/cuda/common.h(778): error:
no suitable user-defined conversion from "__nv_bfloat16" to "__half" exists

return __nv_bfloat162{
    __hfma(a.x, b.x, c.x), __hfma(a.y, b.y, c.y)
};
```

This is not caused by the model using BF16. The first failing generated kernel has only `int*` and `float*` parameters; merely including this TileLang header causes the SM70 compilation failure.

### Minimal reproduction

No model or GPU is required. In a fresh container with TileLang 0.1.9, compile any CUDA file which includes the header, for example:

```cuda
#include <tl_templates/cuda/common.h>

extern "C" __global__ void int_float_only(const int* input, float* output) {
    if (threadIdx.x == 0) output[0] = static_cast<float>(input[0]);
}
```

with the same include roots used by TileLang JIT:

```bash
nvcc --cubin -O3 -lineinfo -arch=sm_70 -std=c++17 \
  -I.../site-packages/tilelang/src \
  -I.../site-packages/tilelang/3rdparty/cutlass/include \
  -o int_float_only.cubin int_float_only.cu
```

It exits with the six `common.h:778` conversion errors above.

### Upstream status and validation

This exact fallback was fixed upstream in TileLang v0.1.10. Compare:

- [v0.1.9 common.h](https://raw.githubusercontent.com/tile-ai/tilelang/v0.1.9/src/tl_templates/cuda/common.h)
- [v0.1.10 common.h](https://raw.githubusercontent.com/tile-ai/tilelang/v0.1.10/src/tl_templates/cuda/common.h)

The newer implementation converts BF16 lanes to float, performs the arithmetic, then converts back to BF16.

We validated this in a separate fresh container by changing only that upstream fallback in the container's writable layer: the identical minimal `sm_70` compilation changed from exit code 1 to exit code 0. No existing service, model mount, or cache was modified.

### Requested fix

Could 1Cat please either:

1. update the compatible TileLang / apache-tvm-ffi dependency pair to a version including the v0.1.10 fix, or
2. cherry-pick the small upstream `common.h` fallback fix into the packaged dependency?

Please also add a small SM70 JIT smoke test that includes `tl_templates/cuda/common.h`, so this path is covered by the V100 release validation.

### Temporary workarounds

- set `VLLM_SM70_FLASHQLA_ORIGINAL_PREFILL=0` to use the compiled VLK varlen implementation, or
- select the Triton GDN prefill backend.

Could you also confirm which GDN prefill backend was used for the official V100 validation of this release?

yangzhuxinyzx:
Fixes #105. Upgrades tilelang and apache-tvm-ffi to 0.1.10, where the SM70 BF16 fallback compiles correctly. Adds an offline nvcc sm_70 header smoke test. Validation: CUDA 12.8 compiles the 0.1.10 header and rejects the 0.1.9 header; the new pytest passes against a temporary 0.1.10 installation.

### DIFF
--- a/requirements/cuda.txt
+++ b/requirements/cuda.txt
@@ -11,8 +11,8 @@ torchvision==0.25.0 # Required for phi3v processor. See https://github.com/pytor
 # FlashInfer should be updated together with the Dockerfile
 flashinfer-python==0.6.11.post2
 flashinfer-cubin==0.6.11.post2
-apache-tvm-ffi==0.1.9
-tilelang==0.1.9
+apache-tvm-ffi==0.1.10
+tilelang==0.1.10
 # Cap nvidia-cudnn-frontend (transitive dep of flashinfer) due to
 # breaking changes in 1.19.0
 nvidia-cudnn-frontend>=1.13.0,<1.19.0
--- a/tests/cuda/test_sm70_tilelang_header.py
+++ b/tests/cuda/test_sm70_tilelang_header.py
@@ -4,12 +4,12 @@
 from __future__ import annotations
 
 import importlib.metadata
-import re
 import shutil
 import subprocess
 from pathlib import Path
 
 import pytest
+import regex as re
 from packaging.version import Version
 
 SM70_TILELANG_FIX_VERSION = Version("0.1.10")
@@ -27,6 +27,7 @@ def test_tilelang_header_compiles_for_sm70(tmp_path: Path) -> None:
     nvcc = shutil.which("nvcc")
     if nvcc is None:
         pytest.skip("requires nvcc")
+    assert nvcc is not None
 
     compiler = subprocess.run(
         [nvcc, "--version"],
@@ -37,6 +38,7 @@ def test_tilelang_header_compiles_for_sm70(tmp_path: Path) -> None:
     match = re.search(r"release (\d+)\.", compiler.stdout)
     if match is None:
         pytest.fail(f"could not determine CUDA version from: {compiler.stdout}")
+    assert match is not None
     if int(match.group(1)) >= 13:
         pytest.skip("CUDA 13 no longer supports SM70 offline compilation")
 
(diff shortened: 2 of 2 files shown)


yangzhuxinyzx:
Closed as superseded. Current main already pins apache-tvm-ffi/tilelang 0.1.10, and the current FlashQLA/TileLang source has moved beyond this old two-file patch. There is no remaining dependency change to merge.

PR #117, opened 2026-08-01
[Bugfix] Protect non-MTP SM70 GDN graph decode (#93)

### LINKED ISSUE #93: [Bug]: FULL-graph GDN decode corruption on non-MTP SM70 Qwen MoE (full-forward guard only auto-arms when speculative_config is set)
### Summary

On v1.2.1 the SM70 Qwen Gated-Delta-Net **full-forward** decode-corruption guard auto-arms **only when `speculative_config` is set (i.e. MTP)**. A **non-MTP** Qwen GDN MoE served under the default SM70 FULL-CUDA-graph compile therefore silently produces garbage decode unless the operator manually exports `VLLM_SM70_QWEN_GDN_FULL_FORWARD=1`.

This is the same corruption class documented in your `SM70_MTP_OUTPUT_QUALITY_AUDIT_20260616.md`, but it is **not** MTP-specific — it also hits non-spec models. Observed on **Qwen3.5-397B-A17B (AWQ), TP8, 8× Tesla V100-SXM2-32GB**.

### Root cause

`QwenGatedDeltaNetAttention` in `vllm/model_executor/layers/mamba/gdn/qwen_gdn_linear_attn.py` gates the auto-arm on `speculative_config is not None`:

```python
self.auto_sm70_qwen_gdn_full_forward = (
    envs.VLLM_SM70_FLASH_V100_0DOT3_COMPILE_GRAPH
    and vllm_config.speculative_config is not None   # <-- excludes all non-MTP models
    and not envs.VLLM_SM70_QWEN_GDN_SPEC_CORE_OP
    and not envs.VLLM_SM70_QWEN_GDN_003_SPEC_CORE_OP
)
```

The FULL-graph capture splits the GDN forward (compiled in-projection → opaque recurrent core → compiled out-projection); the per-step recurrent state is mis-ordered, and the full-forward opaque boundary is the fix. With the `speculative_config` clause, non-MTP models never auto-arm it.

### Reproduce

- Model: Qwen3.5-397B-A17B AWQ (non-MTP), `--tensor-parallel-size 8`
- HW/stack: 8× Tesla V100-SXM2-32GB (SM70), CUDA 12.8 / torch 2.10, 1Cat-vLLM v1.2.1 wheel
- Default SM70 compile-graph (`VLLM_SM70_FLASH_V100_0DOT3_COMPILE_GRAPH`), **without** `VLLM_SM70_QWEN_GDN_FULL_FORWARD`
- Greedy decode → repeated-output collapse / corrupted-token salad.
- Export `VLLM_SM70_QWEN_GDN_FULL_FORWARD=1` → coherent decode. So the guard works; it just isn't armed automatically for non-MTP.

### Suggested fix

Drop the `speculative_config is not None` clause so the guard auto-arms for non-MTP SM70 Qwen GDN compile-graph decode as well (the `force_`/`disable_` overrides are unchanged). One-line change. Verified on 397B TP8: with no env set the worker logs `guard armed (force=False, auto=True)` and decode is coherent (was salad on the FULL graph before).

Patch, for reference, in a downstream 8×V100 fork of your v1.2.1:
https://github.com/dg1kjd/vllm-v100-sxm2-qwen3.5-397b/commit/98706a8d58d74679ff200906ff72a48bf53a44bc

Happy to open a PR against `main` if that's useful. Thanks for the SM70/V100 work — it's what makes running a 397B on this box possible.

yangzhuxinyzx:
Fixes #93.

## Purpose
Arm the existing SM70 Qwen GDN full-forward safety boundary for CUDA-graph decode regardless of whether speculative decoding is enabled. The old automatic condition required MTP, leaving non-MTP FULL graph replay on a known unsafe split recurrent-state path.

## Scope
- Preserve explicit full-forward disable and force overrides.
- Preserve speculative-core and 0.0.3 speculative-core exclusions.
- Keep the deep native MTP 0.0.3 safety fallback armed.

## Validation
- `python -m ruff check vllm/model_executor/layers/mamba/gdn/qwen_gdn_linear_attn.py tests/v1/attention/test_gdn_metadata_builder.py`
- `python -m ruff format --check vllm/model_executor/layers/mamba/gdn/qwen_gdn_linear_attn.py`
- `python -m pytest -q tests/v1/attention/test_gdn_metadata_builder.py -k "full_forward_auto_protects_non_mtp or blocks_003_spec_core_only_for_deep_native_mtp"` (8 passed)

## Hardware gate
The local host has an NVML/driver mismatch, so full V100 graph replay and text-quality validation remain required before this Draft PR is promoted.

### DIFF
--- a/tests/v1/attention/test_gdn_metadata_builder.py
+++ b/tests/v1/attention/test_gdn_metadata_builder.py
@@ -1025,9 +1025,7 @@ def test_sm70_qwen_gdn_full_forward_auto_protects_non_mtp(
 ):
     monkeypatch.setenv("VLLM_SM70_FLASH_V100_0DOT3_COMPILE_GRAPH", "1")
     monkeypatch.setenv("VLLM_SM70_QWEN_GDN_SPEC_CORE_OP", str(int(spec_core)))
-    monkeypatch.setenv(
-        "VLLM_SM70_QWEN_GDN_003_SPEC_CORE_OP", str(int(spec_core_003))
-    )
+    monkeypatch.setenv("VLLM_SM70_QWEN_GDN_003_SPEC_CORE_OP", str(int(spec_core_003)))
     _clear_envs_cache()
 
     assert (
--- a/tests/v1/attention/test_gdn_metadata_builder.py
+++ b/tests/v1/attention/test_gdn_metadata_builder.py
@@ -928,7 +928,7 @@ def test_spec_commit_pure_decode_consumes_padded_graph_rows_without_metadata_pat
     assert live_meta.num_accepted_tokens is not padded_accepted
 
 
-def test_sm70_qwen_gdn_full_forward_auto_is_mtp_engine_scoped(
+def test_sm70_qwen_gdn_full_forward_guard_honors_auto_flag(
     monkeypatch,
     local_gdn_model,
 ):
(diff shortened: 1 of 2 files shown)


yangzhuxinyzx:
Closed after the current-main output-quality audit. The no-MTP instability was localized to nondeterministic FP16 atomic accumulation in the AWQ MoE weighted reduction and fixed in #225; with that fix, three matched no-MTP greedy requests are token-exact while retaining the lightweight GDN path. This PR instead broadens the heavier full-forward GDN guard to every no-MTP graph engine, without a matched speed/quality A/B proving that boundary is still needed. Its root-cause attribution is therefore superseded and the change would add an unproven performance tax.

PR #120, opened 2026-08-01
[Bugfix] Extend SM70 non-MTP CUDA graph coverage (#89)

### LINKED ISSUE #89: [Usage]: 单并发很快很准 任务数量在3-6时就每秒1t了 很慢
### Your current environment

```text
The output of `python collect_env.py`
```

### How would you like to use vllm

并发问题
单并发很快很准 任务数量在3-6时就每秒1t了 很慢

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

yangzhuxinyzx:
Fixes #89.

## Root cause
The SM70 Flash-V100 compile-graph policy hard-coded non-MTP capture sizes to `[1, 2]`. Any decode batch of three or more requests missed CUDA graphs and fell onto a much slower fallback. This branch also bypassed the existing `VLLM_SM70_DENSE_CUDAGRAPH_CAPTURE` control, making the documented opt-in ineffective for this route.

## Change
- Default non-MTP SM70 graph capture now grows by powers of two only through batch 8, preserving single-request startup and graph-memory behavior while covering modest serving concurrency.
- `VLLM_SM70_DENSE_CUDAGRAPH_CAPTURE=1` now takes effect for this compile-graph policy and captures `1..min(max_num_seqs, 64)`.
- Explicit user `cudagraph_capture_sizes` remain untouched. MTP-specific shape logic is unchanged.

## Validation
- `python -m ruff check vllm/config/vllm.py tests/test_config.py`
- `python -m ruff format --check vllm/config/vllm.py tests/test_config.py`
- `python -m pytest -q tests/test_config.py -k sm70_non_spec_cudagraph_capture_sizes` (7 passed)

## Hardware gate
Needs a V100 non-MTP concurrency sweep to record actual graph route hits, capture startup cost, and TPOT for batch 3-8 before promotion.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -32,6 +32,7 @@ from vllm.config.utils import get_field
 from vllm.config.vllm import (
     OPTIMIZATION_LEVEL_TO_CONFIG,
     OptimizationLevel,
+    _sm70_non_spec_cudagraph_capture_sizes,
 )
 from vllm.platforms import current_platform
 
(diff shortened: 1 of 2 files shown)


Leonccaa:
Independent B=3 datapoint that may help choose the bounded default:

On Qwen3.6-27B AWQ / TP4 / MTP0 with `max_num_seqs=3`, the prior SM70 Flash-V100 `[1, 2]` policy left B=3 on the dynamic fallback at **41.639 aggregate decode tok/s**. Adding an exact B=3 CUDA graph raised the base path to **80.467 tok/s** (+93%). This indicates that the dominant effect in this experiment was CUDA Graph coverage.

This is only an endpoint measurement for B=3; it does not compare exact-3 directly against padded-to-4, and we have not measured B=4-8. Still, it suggests considering `[1, 2, 3, 4, 8]` for the bounded default while retaining the existing dense opt-in for deployments that need finer-grained capture.

yangzhuxinyzx:
Closed as superseded by current main. The current helper `_sm70_nomtp_cudagraph_capture_sizes` already covers bounded non-MTP shapes `[1,2,4,8,16]`, includes the exact configured max request count, and is wired into the SM70 compile-graph policy. This older patch conflicts with that newer implementation and adds no remaining independent behavior. Larger-batch Flash-V100 graph support still needs its own physical route gate; it should not be inferred from this stale config-only PR.

PR #122, opened 2026-08-01
[Bugfix] Support 1.x hidden metrics version checks

yangzhuxinyzx:
## Root cause
`_prev_minor_version_was()` asserted that every release had major version zero. 1Cat-vLLM has 1.x releases, so `--show-hidden-metrics-for-version` could raise an assertion instead of evaluating the requested previous minor version.

## Change
Support any nonzero major version and return false for an x.0 release, where no same-major previous minor exists.

## Validation
- `python -m ruff check vllm/version.py tests/test_version.py`
- `python -m ruff format --check vllm/version.py tests/test_version.py`
- `python -m pytest -q tests/test_version.py` (13 passed)


yangzhuxinyzx:
CI note: the hosted full pre-commit run fails on the pre-existing repository baseline tracked in #126, including all-files formatting, typos, Markdownlint, SPDX, pip-compile, and mypy debt. This PR does not touch those failing baseline files. Local validation: tests/test_version.py passed with 13 tests. Keep Draft until #126 is resolved, then rerun hosted CI.

PR #123, opened 2026-08-01
[Bugfix] Align SM70 paged prefill accumulators (#98)

### LINKED ISSUE #98: [Bug]: bundled flash_attn_v100 1.2.0 paged prefill kernels are nondeterministic — run-to-run |diff| up to 0.19 on identical inputs (official v1.2.1 wheel)
### Environment
- Official wheel `1cat_vllm-1.2.1-cp312-cp312-linux_x86_64.whl`, installed exactly per the README (`--extra-index-url https://download.pytorch.org/whl/cu128`) → torch 2.10.0+cu128
- Tesla V100-SXM2-32GB (SM70), driver 580.126.09, Python 3.12
- Bundled `flash_attn_v100` reports version 1.2.0

### What we found
Kernel-level probe of the public paged APIs: **identical input tensors, 6 repeated calls**, measuring max run-to-run |diff| of the outputs (fp16):

| API | shape | max run-to-run \|diff\| |
|---|---|---|
| `flash_attn_prefill_paged` | block=256, L=300, HQ/HK=12/2, D=128 | **0.1917** |
| `flash_attn_prefill_paged` | block=256, L=512 (full pages) | **0.0999** |
| `flash_attn_prefill_paged` | block=16, L=300 | **0.1575** |
| `flash_attn_decode_paged` | same protocol | 0.000000 (bit-exact) |

Against an fp32 reference the prefill outputs intermittently show max_abs excursions of 0.05–0.11 (cos ≥ 0.99998) on any page size **including fully-aligned pages** — so this is not a paged-addressing problem. (We specifically audited page-straddle + garbage-tail + permuted-block-table addressing on both prefill and decode: all correct, cos=1.000000 on the good runs. Nice work on that, by the way — we went in looking for an addressing bug and found none.)

### Repro
```python
import torch
from flash_attn_v100 import flash_attn_prefill_paged
torch.manual_seed(0)
DEV, DT = "cuda", torch.float16
HQ, HK, D, L, block = 12, 2, 128, 300, 256
nb = (L + block - 1) // block
k = torch.randn(L, HK, D, device=DEV, dtype=DT); v = torch.randn(L, HK, D, device=DEV, dtype=DT)
kc = torch.full((nb, block, HK, D), 100.0, device=DEV, dtype=DT); vc = kc.clone()
for t in range(L):
    kc[t // block, t % block] = k[t]; vc[t // block, t % block] = v[t]
bt = torch.tensor([list(range(nb))], dtype=torch.int32, device=DEV)
sl = torch.tensor([L], dtype=torch.int32, device=DEV)
q = torch.randn(1, L, HQ, D, device=DEV, dtype=DT)
outs = [flash_attn_prefill_paged(q, kc, vc, bt, sl, causal=True).clone() for _ in range(6)]
print("max run-to-run |diff|:", max((a - b).abs().max().item() for a in outs for b in outs))
# observed on V100: up to ~0.19; expected for a deterministic kernel: 0.0
```

### Hypothesis
The extension exports `prefill_paged_splitkv_fwd` — the signature looks like a split-KV reduction race: non-fixed accumulation order (or a missing sync) when partial results are combined. Decode being bit-exact under the same protocol supports this.

### Impact
Silent run-to-run output variation at serving level (temperature-0 responses can differ across identical requests). To be clear: in our end-to-end serving tests (Qwen2.5-0.5B, Qwen2.5-14B-AWQ TP4, gemma-4-E2B/31B) we saw zero NaN/degenerate outputs and excellent performance — this is a determinism/precision-jitter report, not a correctness-catastrophe one.

### Ask
Is the split-KV prefill path known/expected to be nondeterministic? If yes, a doc note would help; a deterministic reduction (fixed combine order) would be better for temperature-0 reproducibility.

---
Found during an evaluation of V100 vLLM stacks (AI-assisted test harness; all results verified on physical hardware — DGX-1, 8× V100-SXM2-32GB). Full probe harness available on request.

yangzhuxinyzx:
## Purpose
Fixes #98: the v1.2.1 D=128 paged-prefill shared accumulator started at byte offset 80,144 (16-byte aligned) while WMMA load/store require 256-bit alignment. This is undefined behavior and explains the physical-V100 replay jitter.

## Implementation
- Align the paged-prefill accumulator buffer and dynamic shared-memory base to the WMMA contract.
- Add a compile-time layout assertion.
- Add a physical-SM70 regression for six synchronized D=128 paged-prefill replays at L=300 and L=512, checking both the direct extension and public wrapper.

## Validation
- `CUDA_HOME=/usr/local/cuda TORCH_CUDA_ARCH_LIST=7.0 MAX_JOBS=1 python setup.py build_ext ...`: passed; produced SM70 cubins and imported the extension.
- `ruff check` / `ruff format --check` for the new test: passed.
- Target pytest: skipped locally because the host CUDA driver is unavailable (`cudaGetDeviceCount` error 804).

## Required Hardware Gate Before Merge
Run `pytest -q tests/kernels/attention/test_sm70_flash_v100_paged_prefill_determinism.py` on a physical V100 with the rebuilt extension. Both direct and public APIs must be bit-exact across all six replays; also rerun an affected temperature-0 serving reproducibility probe.

Base: `3ec0c68c6596d6ab31fbdee9fa676254a52c2b7d`
Head: `e9a7f9aa6`

### DIFF
--- a/flash-attention-v100/kernel/fused_mha_forward_paged.cu
+++ b/flash-attention-v100/kernel/fused_mha_forward_paged.cu
@@ -3,6 +3,7 @@
 #include <cuda_fp16.h>
 #include <torch/extension.h>
 #include <algorithm>
+#include <cstddef>
 #include <cstdlib>
 #include <cstring>
 #include <string>
@@ -135,7 +136,8 @@ struct KernelConfig {
         alignas(16) float  pipeline_s[PIPELINE_S_ELEMENTS];
         alignas(16) int    pipeline_page_idx[PIPELINE_PAGE_ELEMENTS];
         alignas(16) int    pipeline_page_offset[PIPELINE_PAGE_ELEMENTS];
-        alignas(16) float  o      [BLOCK_M * O_STRIDE];
+        // WMMA accumulator loads and stores require 256-bit alignment.
+        alignas(32) float  o      [BLOCK_M * O_STRIDE];
         alignas(16) float  row_max[BLOCK_M];
         alignas(16) float  row_sum[BLOCK_M];
         alignas(16) int    page_idx[LOW_SMEM_PAGE_COUNT];
@@ -143,6 +145,8 @@ struct KernelConfig {
     };
 
     static constexpr size_t TOTAL_SMEM = ((sizeof(SmemLayout) + 127) & ~size_t(127));
+    static_assert(offsetof(SmemLayout, o) % 32 == 0,
+                  "WMMA accumulator storage must be 256-bit aligned");
 };
 
 template<typename Config>
@@ -480,7 +484,7 @@ flash_attention_forward_kernel_paged(
     const int64_t k_row_stride = k_token_stride;
     const int64_t v_row_stride = v_token_stride;
 
-    extern __shared__ char smem_raw[];
+    extern __shared__ __align__(128) char smem_raw[];
     init_smem<Config>(smem_raw);
     auto& smem = *reinterpret_cast<typename Config::SmemLayout*>(smem_raw);
 
--- a/tests/kernels/attention/test_sm70_flash_v100_paged_prefill_determinism.py
+++ b/tests/kernels/attention/test_sm70_flash_v100_paged_prefill_determinism.py
@@ -1,0 +1,129 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+from __future__ import annotations
+
+import pytest
+import torch
+
+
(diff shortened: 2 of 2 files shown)


yangzhuxinyzx:
Superseded by #226, which ports this fix onto current main, updates the regression to the current accelerator API, and completes the previously missing physical-V100 gate. The rebuilt current-source extension passed both L=300 and L=512 six-replay tests bitwise, plus an independent FP32 reference check. #226 merged at f2a7937c7061e24454e7c4650f4036188033bbb4.

PR #124, opened 2026-08-01
[Bugfix] Support inline Anthropic system messages (#86)

### LINKED ISSUE #86: [Bug]: [Upstream Sync] Support system-role messages inside Anthropic Messages API messages array (follow vllm-project/vllm#44283)
### Your current environment

This problem is environment unrelated so no env report needed.

### 🐛 Describe the bug

Claude Code CLI >= 2.1.154 (and Claude Desktop Code mode) started sending `system` role messages inside the `messages` array of the Anthropic Messages API, instead of (or in addition to) the top-level `system` field.

Our fork's Anthropic protocol still strictly validates `message.role` as `Literal["user", "assistant"]`, which causes every request from recent Claude Code versions to fail with HTTP 400 before any inference happens.

This is the same issue reported and fixed upstream in:

- **Upstream issue:** [vllm-project/vllm#44000](https://github.com/vllm-project/vllm/issues/44000) 
- **Upstream fix:** [vllm-project/vllm#44283](https://github.com/vllm-project/vllm/pull/44283) — *Support system role messages inside messages array* 

The fix has been merged into upstream release **v0.23.0** .

**Example error:**

```text
API Error: 400 {
  "error": {
    "message": "1 validation error:\n  {'type': 'literal_error', 'loc': ('body', 'messages', 1, 'role'), 'msg': \"Input should be 'user' or 'assistant'\", 'input': 'system', 'ctx': {'expected': \"'user' or 'assistant'\"}}",
    "type": "Bad Request",
    "param": null,
    "code": 400
  }
}
```

---

### Steps to reproduce

1. Start vLLM from this fork with the Anthropic-compatible `/v1/messages` endpoint.
2. Configure Claude Code (>= 2.1.154) or Claude Desktop to use the local endpoint via `ANTHROPIC_BASE_URL`.
3. Send any prompt.
4. vLLM returns HTTP 400 due to `role` validation failure.

---

### Expected behavior

- Fork should remain compatible with Claude Code >= 2.1.154.
- `system` role messages inside `messages[]` should be normalized into the top-level Anthropic `system` field (or otherwise handled gracefully), matching upstream behavior.
- Requests should not be rejected before model inference.

---

### Actual behavior

Every request from Claude Code >= 2.1.154 fails with 400 Bad Request due to strict role validation.

---

### Suggested fix / upstream reference

Please cherry-pick or port the upstream fix from **PR #44283** (merged into v0.23.0). The upstream fix adds support for `system` role messages inside the `messages` array in the Anthropic protocol layer, rather than simply expanding the allowed enum.

Key files likely involved (upstream):
- `vllm/entrypoints/anthropic/protocol.py` — role validation
- `vllm/entrypoints/anthropic/` — request normalization logic

If a full sync is not feasible, a minimal backport of the `system`-in-messages normalization would resolve the immediate compatibility issue.

---

### Workaround (for users before fix)

Downgrade Claude Code to **v2.1.153** (last version before the breaking change) and disable auto-update by setting `"autoUpdate": false` in `~/.config/claude/settings.json`.

---

### Additional context

- This affects both Claude Code CLI and Claude Desktop's Code mode when routed through the Anthropic `/v1/messages` path.
- The OpenAI-compatible `/v1/chat/completions` path does not exhibit this exact validation error.
- Multiple downstream forks and platforms have already backported this fix (e.g., vllm-ascend, higress, xorbitsai/inference).

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

yangzhuxinyzx:
## Purpose
Fixes #86: current Anthropic `/v1/messages` validation rejects `role: system` entries inside `messages`, which recent Claude Code clients send.

## Implementation
Exact backport of upstream vLLM #44283 (`ed9a7526b6deea866c04edcaacbbecadb591b7ef`):
- accept `system` in the Anthropic message schema;
- merge top-level and inline system text into one leading OpenAI system message;
- omit inline system entries from normal conversation conversion;
- preserve the billing-header stripping behavior.

## Validation
- `ruff check` and `ruff format --check`: passed.
- `PYTHONPATH=<worktree> conda run -n vllm-0.0.5-t210 python -m pytest -q tests/entrypoints/anthropic/test_anthropic_messages_conversion.py`: **31 passed**.
- Direct `AnthropicCountTokensRequest` inline-system conversion smoke: passed.

Base: `3ec0c68c6596d6ab31fbdee9fa676254a52c2b7d`
Head: `5ad3ddc5f`


yangzhuxinyzx:
CI note: the hosted full pre-commit run fails on the pre-existing repository baseline tracked in #126, including all-files formatting, typos, Markdownlint, SPDX, pip-compile, and mypy debt. This PR does not touch those failing baseline files. Local validation: Anthropic conversion tests passed with 31 tests. Keep Draft until #126 is resolved, then rerun hosted CI.

Issue #126, opened 2026-08-01
[CI/Build] Restore a green full pre-commit baseline on main


yangzhuxinyzx:
Current main baseline fails the repository full pre-commit workflow and blocks otherwise validated bug-fix PRs.

Acceptance: a clean worktree at main passes `pre-commit run --all-files --hook-stage manual` with no modifications or failures. Mechanical cleanup and semantic type/dependency fixes remain in separate reviewable Draft PRs. No V100 performance or kernel behavior change is included in this tracking issue.

Progress:
- [x] Requirements resolver conflict: PR #127, reviewed locally.
- [x] Python Ruff formatting: PR #128, reviewed locally.
- [x] Markdownlint: PR #129, reviewed locally.
- [x] Python SPDX headers: PR #130, reviewed locally.
- [x] Existing C/CUDA typos, formatting, and C/CUDA SPDX scope: reviewed locally in PRs #134, #135, #136, #139, and #140. PRs #131 and #133 were superseded without merge.
- [x] Remaining formatter artifacts split for review: C/CUDA microbenchmarks #143, FlashInfer Volta header #145, Docker dependency graph #142.
- [x] Forbidden standard-library `re` imports split for review: benchmark scripts #144 and the TileLang CUDA test follow-up in #116.
- [x] Mypy baseline and behavior coverage: #132 now includes MTP graph-variant regression tests; #141 fixes the final GDN Py3.13 annotation.
- [x] Scoped ShellCheck findings: #146.
- [x] `torch.cuda` API policy baseline: #147 through #158 preserve synchronization/capability semantics in source, tests, and benchmarks.
- [x] Clean integration worktree passed `pre-commit run --all-files --hook-stage manual` locally on 2026-08-01, with no modifications.
- [ ] Run hosted checks after reviewed Draft PRs are promoted and merged; V100-specific correctness/performance gates remain outside this static-baseline tracker.

yangzhuxinyzx:
Mypy baseline repair is ready for review in Draft PR #132: https://github.com/1CatAI/1Cat-vLLM/pull/132. Validation: Python 3.10 through 3.12 mypy passed, changed groups pass under Python 3.13, ruff check passed, and the focused pytest suite passed 64 tests. The remaining Python 3.13 errors are five pre-existing has-type diagnostics in vllm/v1/attention/backends/gdn_attn.py and are outside this PR scope.

PR #148, opened 2026-08-01
[Core] Preserve static draft vocab synchronization

yangzhuxinyzx:
## Purpose

Resolve the #126 check-torch-cuda-call finding in static/dynamic draft-vocabulary initialization without changing synchronization behavior.

## Base

- onecat/main@3ec0c68c6596d6ab31fbdee9fa676254a52c2b7d

## Change

Only vllm/v1/spec_decode/static_draft_vocab.py changes.

- Keep device synchronization immediately after TP all_to_all_single in initialize_static_draft_vocab.
- Keep device synchronization immediately after _refresh_gpu_tail / _refresh_tail in initialize_dynamic_draft_vocab.
- Preserve the same device argument, ordering, and CUDA-backed execution path; replace only torch.cuda.synchronize(device) with torch.accelerator.synchronize(device).

## Validation

- python -m pytest -q tests/v1/spec_decode/test_static_draft_vocab.py: 18 passed.
- check-torch-cuda-call on the changed file: passed.
- Ruff check and Ruff format --check on the changed file: passed.
- git diff --check: passed.
- File-scoped default pre-commit hooks pass except five existing mypy errors at unchanged lines 421-436 in this file. They are outside this two-line diff and remain tracked separately under #126.

## V100 Gates Still Required

Run on real V100 hardware after integration:

- TP2 and TP4 static/dynamic draft-vocabulary initialization.
- MTP CUDA-graph replay, acceptance length, and long-output quality for AWQ and FP8 routes.
- Confirm synchronization ordering around TP communication and dynamic-tail refresh; no performance claim is made here.

### DIFF
--- a/vllm/v1/spec_decode/static_draft_vocab.py
+++ b/vllm/v1/spec_decode/static_draft_vocab.py
@@ -782,7 +782,7 @@ def initialize_static_draft_vocab(
             input_split_sizes=input_splits,
             group=tp_group.device_group,
         )
-    torch.cuda.synchronize(device)
+    torch.accelerator.synchronize(device)
 
     fingerprints = payload.get("fingerprints")
     ranking_fingerprint = None
@@ -1044,5 +1044,5 @@ def initialize_dynamic_draft_vocab(
         )
     else:
         runtime._refresh_tail()
-    torch.cuda.synchronize(device)
+    torch.accelerator.synchronize(device)
     return runtime


yangzhuxinyzx:
Closed as superseded. Current main already uses torch.accelerator.synchronize(device) at both static and dynamic draft-vocabulary initialization boundaries.

PR #128, opened 2026-08-01
[CI/Build] Clear Python Ruff format baseline (#126)

yangzhuxinyzx:
## Purpose
Clear only the existing Python Ruff format baseline debt tracked by #126.

## Scope
- Applied the repository-pinned Ruff 0.14.0 formatter to 74 tracked Python files.
- Excluded all 15 files in the concurrent mypy worker write scope.
- No manual or semantic edits; no docs, requirements, C/CUDA, or Markdown changes.

## Duplicate Check
No open PR was found for Ruff format baseline cleanup. This is a bounded mechanical child of #126.

## Validation
- Ruff 0.14.0 format check passed for all 3,421 eligible Python targets.
- The repository ruff-format pre-commit hook passed for all 74 changed files.
- Python AST equivalence against HEAD passed for all 74 changed files.
- Git diff check passed; scope assertions confirmed no excluded or non-Python files changed.

## Runtime Validation
No physical GPU is available. This formatter-only PR makes no GPU or performance claim.

## Remaining Risk
Full pre-commit remains tracked by #126; other formatter, lint, type, and dependency baseline items are intentionally outside this PR.

## AI Assistance
Prepared with OpenAI Codex; human review is required before merge.

PR #152, opened 2026-08-01
[CI/Build] Use accelerator warmup synchronization

yangzhuxinyzx:
## Purpose

Part of #126. Replace the SM70 TurboMind warmup synchronization with `torch.accelerator.synchronize(device)` while preserving its exact location, explicit device, warmup ordering, and error behavior.

## Scope

Only `vllm/model_executor/warmup/awq_sm70_warmup.py` changes: one line.

## Test Plan

- `check-torch-cuda-call` on the changed file
- Ruff check and Python compile
- CPU mock-worker ordering test for AWQ/FP8/FP4/MoE warmup -> synchronize(device) -> LUT export
- `git diff --check`

## Test Result

All scoped checks above pass. Mypy 3.12/3.13 retain the pre-existing line 593 `worker.device` optionality error; this PR does not modify that line. Ruff format reports six existing formatting differences already covered by #128; no formatting diff is included here.

## Required V100 Gate

Before merge, validate SM70 worker startup with TurboMind AWQ, FP8, and NVFP4 paths, including failure propagation and LUT import/export behavior.

### DIFF
--- a/vllm/model_executor/warmup/awq_sm70_warmup.py
+++ b/vllm/model_executor/warmup/awq_sm70_warmup.py
@@ -641,7 +641,7 @@ def sm70_awq_warmup(worker: Worker) -> None:
             moe_layers, moe_token_counts
         )
         single_token_calls = _warmup_moe_single_token_layers(moe_layers)
-    torch.cuda.synchronize(device)
+    torch.accelerator.synchronize(device)
     logger.info(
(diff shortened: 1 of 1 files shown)


yangzhuxinyzx:
Closed as superseded. Current main already uses torch.accelerator.synchronize(device) at the SM70 warmup boundary (and at the newer coordinated FP8 tuning boundaries), so this one-line patch has no remaining delta.

PR #166, opened 2026-08-02
[SM70][NVFP4] ModelOpt admit + hybrid + MoE fallback + W4A4 + Medium scale fix

rivetphilbot:
## Summary

Enables **NVFP4 on Tesla V100 (SM70)** for Qwen3.6-class models:

1. **ModelOpt W4A16 NVFP4 admit** — `get_min_capability` **75 → 70**; wire through `sm70_turbomind`.
2. **Hybrid linear attention policy** — GDN `linear_attn` linears; default `tm`, stamp + env override, dequant path.
3. **W4A4-on-disk weight path** — CT/ModelOpt weight-only / W4A16 semantics; acts remain half (logged).
4. **MoE NVFP4 SM70 compat** — auto-select prefers **Marlin then EMULATION**.
5. **Medium/TC scale convention fix** — tiny block-scale GPTQ exports no longer under-scale through TM (`block * 1/G`); detect `max|weight_scale| < 1` and skip global factor. Fixes token salad on Medium/TC NVFP4 while keeping Aggressive path unchanged.

### Medium/TC NVFP4 (important)

Field check: Aggressive and Medium share the same CT schema, but **block-scale ranges differ** (Aggressive ~O(10–448), Medium GPTQ ~O(1e-3)). SM70 prepare does `block * global` after CT stores `global=1/disk_global`, so Medium was under-scaled by ~G. Load-time convention detect fixes serve without requant.

Validated on pve3: `grok-medium2-nvfp4` equivalent fix → **Paris** + **17×19=323**.

### Hybrid policy

Env `VLLM_SM70_NVFP4_LINEAR_ATTN` → stamp `quantization_config.sm70_linear_attn` → default **`tm`**.

### MoE SM70

Marlin → EMULATION first on Volta. Not a native `nvfp4_moe_*` kernel.

## Serve (dense)

```bash
export VLLM_SM70_FLASH_ATTN_V100=1
export VLLM_SM70_QUANT_BACKEND=turbomind
export VLLM_SM70_NVFP4_TURBOMIND=1
# Medium/TC NVFP4: no special env needed after scale fix
python -m vllm.entrypoints.openai.api_server \
  --model <nvfp4-checkpoint> --trust-remote-code --dtype half \
  --attention-backend FLASH_ATTN_V100 -tp 2 --max-num-seqs 1 \
  --kv-cache-dtype fp8_e5m2
```

## Tests

- `tests/quantization/test_sm70_nvfp4_scale_convention.py` (combine helpers)
- `tests/quantization/test_sm70_nvfp4_moe_fallback.py` (MoE + W4A4 routing)
- pve3: 14 passed; live medium2 Paris/math smoke with scale fix

## Out of scope

- True W4A4 act-quant on SM70
- Native `nvfp4_moe_*` SASS kernels
- Changing quant export to unify CT conventions at source (optional follow-up)

### DIFF
--- a/docs/design/sm70_nvfp4_checkpoint_serve_runbook.md
+++ b/docs/design/sm70_nvfp4_checkpoint_serve_runbook.md
@@ -113,6 +113,10 @@ python3 scripts/sm70_nvfp4_smoke_paris.py --model my-nvfp4
 
 ## Graph vs eager bisect
 
+pve3 A/B on `tc-medium-nvfp4` (2026-08-03): **graph PASS**, **eager PASS**
+(Paris + 323 both modes) with scale normalize + campaign flags. The graph-safe
+compilation/spec pins above are the preferred production path.
+
 ```bash
 bash scripts/sm70_nvfp4_ab_graph_eager.sh /path/to/ckpt served-name
 ```
--- a/docs/design/sm70_nvfp4_checkpoint_serve_runbook.md
+++ b/docs/design/sm70_nvfp4_checkpoint_serve_runbook.md
@@ -1,0 +1,132 @@
+# SM70 NVFP4 checkpoint scale + serve runbook
+
+Date: 2026-08-03  
+Related: PR scale normalize (`normalize_nvfp4_global_scale_for_sm70`),
+`docs/design/sm70_nvfp4_turbomind_operator_optimization.md`
+
+## Problem summary
+
+### 1) Scale convention (Python fix in this PR)
+
+Some GPTQ / Magpie NVFP4 exports (Medium, ThinkingCap, cyber) store **tiny**
+FP8 block scales with a large disk global. CT turns global into `1/G`;
+TurboMind multiplies `block * global` and under-scales → token salad.
+
+**Fix:** detect tiny block scales and set `weight_global_scale = 1.0`
+(leave block scales untouched).
+
+### 2) CUDA-graph garbage with identical wheel (serve flags)
+
+Third-party V100 hosts with the **same** public 1.2.2 `_C.abi3.so` have seen
+coherent output under `--enforce-eager` but salad under default graphs.
+
+**Fix (preferred over permanent eager):** pin graph capture mode and greedy
+MTP draft sampling:
+
+```bash
+  --compilation-config '{"cudagraph_mode":"full_and_piecewise","cudagraph_capture_sizes":[1,2,4,8]}' \
+  --speculative-config '{"method":"mtp","num_speculative_tokens":4,"attention_backend":"FLASH_ATTN_V100","draft_sample_method":"greedy"}'
+```
+
+Why these matter on 1.2.2:
+
+- **`full_and_piecewise` + explicit capture sizes** — avoids the
+  “MTP accepting but not amortized / wrong graph partition” failure mode
+  (piecewise-only or missing size-1 graphs). See deckard tok/s runbook notes.
+- **`draft_sample_method: greedy`** — 1.2.2 auto-promotes MTP draft to
+  probabilistic, which requires `draft_probs` only the dynamic-vocab path
+  emits. With dynamic draft vocab off (or multi-seq), non-greedy draft can
+  corrupt or crash; greedy is the safe production pin (same as
+  voice-coexist-mtp122).
+
+Also set `VLLM_SM70_MTP_DYNAMIC_DRAFT_VOCAB_DEFAULT=0` when
+`max_num_seqs != 1`.
+
+## Wheel identity
+
+Public **1Cat-vLLM 1.2.2** release:
+
+| Artifact | sha256 |
+| --- | --- |
+| wheel | `8a628983ad9d675559910372643220c418b307ddc7fd52ac65a7f5fbcb104bc6` |
+| `_C.abi3.so` | `ffc6271aaf25d96fe690d0f899544801cc1e39486905c49ad0090cb2cbe7a147` |
+
+https://github.com/1CatAI/1Cat-vLLM/releases/tag/v1.2.2  
+Tag commit: `644d8a7cd05ed4ecd1cd188e3c05b4bbd074f504`
+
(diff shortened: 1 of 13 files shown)


rivetphilbot (commented):
## Review — the `auto` default is blind; recommend "stamp the policy on the checkpoint"

Structure is right: ModelOpt admit + the hybrid `linear_attn` split is the correct shape, and the prepare/apply pairing is sound (`try_apply_sm70_nvfp4_linear` → `has_dequant_linear` → `apply_dequant_linear`). One gap worth fixing before merge, plus a root-cause thread.

### The `auto` scale-health heuristic can't see the failure (empirical)

I ran `nvfp4_scale_health_ok`'s exact logic against a **known-garbage** Medium NVFP4 GDN linear and the **known-good** Aggressive one (`in_proj_qkv`/`in_proj_z`/`out_proj`, layer 10):

```
MED2 (garbages on TM):  gmax=2736  gmin=2.7e3  block-ratio=8.0e1  => PASS
AGG  (correct on TM):   gmax=2736  gmin=2.7e3  block-ratio=2.0e2  => PASS
```

Both PASS — the scale stats are nearly identical (block ratios 80 vs 199, both orders of magnitude under the 1e6 gate). So in `auto` (**the default**), Medium/TC route to TurboMind and still garbage. The signal the default rests on doesn't discriminate broken from safe.

That makes `auto` the worst shape: TM + false confidence on fragile checkpoints, and sometimes needless dequant on good ones — speed *and* safety lose.

### Recommend: default `tm`, and stamp the policy on the checkpoint

Fragility is a **per-checkpoint property known at build/validate time**, not a runtime-guessable one. Put the decision where the knowledge is:

1. **Default `tm`.** Aggressive already proves hybrid-on-TM is fast and correct — don't tax the good case, and don't make ops remember `=tm`.
2. **Fragile checkpoints self-declare** — stamp e.g. `quantization_config.sm70_linear_attn: "dequant"` at quant time and have the loader read it. Medium/TC then get the safe path with **zero env to forget**; Aggressive stays on TM. `VLLM_SM70_NVFP4_LINEAR_ATTN` becomes an *override*, not the primary switch.
3. **Drop `auto`** (or ship it off by default). A heuristic that can't tell safe from broken is strictly worse than an explicit stamp; false confidence is the dangerous failure mode.

This gives speed by default and safety by construction rather than by guess.

### Root-cause worth chasing (retires the dequant crutch)

Med2 and Agg have near-identical scale stats yet one garbages on TM and one doesn't — so the fragility is **not in the scales**, it's in how the TM kernel handles those specific packed weights. If that's found, fragile checkpoints could be made TM-safe at build time and nobody pays the dequant tax on those fat `in_proj_qkv` layers at all. Stamped-dequant is the correct interim; this is the real win.

### Minor
- `prepare_nvfp4_dequant_linear` hardcodes `fp16` → bf16 models eat a per-forward `.to(x.dtype)` in `apply_dequant_linear` plus a precision round-trip. Dequant to the layer's compute dtype.
- `weight_global_scale` isn't freed after dequant (tiny, but `weight`/`weight_scale` are).
- No correctness test comparing dequant output vs a reference on a known-garbage layer — a golden test there would have caught the `auto` blindness in CI.

rivetphilbot (commented):
Implemented the review direction on this branch:

1. **Default is now `tm`** — hybrid GDN candidates use TurboMind unless told otherwise (no permanent dequant tax on Aggressive).
2. **Checkpoint stamp** — loader reads `quantization_config.sm70_linear_attn` (`tm`|`dequant`|`auto`). Fragile Medium/TC builds should stamp `"dequant"` at quant/validate time so ops need no env.
3. **Resolution order** — explicit `VLLM_SM70_NVFP4_LINEAR_ATTN` env **overrides** stamp → stamp → default `tm`.
4. **`auto` retained only as an explicit/experimental choice** (not the default); scale-health is documented as not a quality oracle.
5. Minors: dequant uses model/layer compute dtype (not hardcoded fp16); `weight_global_scale` freed after dequant.

Root-cause of TM vs packed weights (why Med2 garbages with the same scale stats as Agg) remains follow-up — stamped dequant is the interim.

Quant-side stamp example for fragile builds:
```json
"quantization_config": {
  "...": "...",
  "sm70_linear_attn": "dequant"
}
```

yangzhuxinyzx:
Closing this combined PR after a current-main output-quality audit. It is not safe to merge as written.

Two blocking correctness findings:

1. The scale convention is inferred from `max(abs(weight_scale)) < 1`. That value cannot identify the on-disk convention: block and global scales are multiplicatively non-identifiable without checkpoint metadata. Running the PR's actual helper on a valid CT-style encoding with an FP8-representable block scale of `0.5` and disk global `100` changes the expected dequantized value `2 * 0.5 / 100 = 0.01` to `2 * 0.5 * 1 = 1.0`, a 100x error. The added tests encode the heuristic's assumed answer but never compare against an independent checkpoint-format/reference dequant oracle.

2. `normalize_nvfp4_global_scale_for_sm70()` is called unconditionally in both CT W4A16 and W4A4 `process_weights_after_loading()` before `try_prepare_sm70_nvfp4_linear()` checks the device. Therefore the heuristic rewrites scales for non-SM70 Marlin/CUTLASS paths too; this is not scoped to Volta despite the helper name.

The PR is also conflicting with current main. Main now already includes the generic Marlin/EMULATION NVFP4 MoE candidates and substantially newer SM70 NVFP4 routing/quality work, including current-source TP4 output-health evidence. The Paris/323 smoke and graph/MTP pins here are not sufficient numerical or model-quality evidence for a format-level scale rewrite.

A still-useful piece, ModelOpt NVFP4 admission on exact SM70, should be resubmitted as a small standalone PR only after it has: (a) exact-SM70-only routing, (b) explicit checkpoint format/convention provenance rather than a magnitude threshold, (c) real-layer dequant/logit comparison against an independent reference, and (d) current CUDA-graph no-MTP/MTP output-quality coverage as applicable.

PR #167, opened 2026-08-02
[Kernel] Screen SM70 TP8 small all-reduce

yangzhuxinyzx:
## Purpose

Screen a default-off custom all-reduce route for DeepSeek-V4-Flash TP8 decode on the non-fully-connected 8x V100 topology. The trace contains 87 FP16 8 KiB reductions per token.

## Test Plan

- Validate the topology/architecture gate with unit tests.
- Compare graph-captured rank-max latency and correctness against PyNCCL for 4096 FP16 elements.
- Keep the route only if projected savings exceed 0.2 ms/token and the later accumulated endpoint quality gate passes.

## Test Result

- `tests/distributed/test_sm70_nonfull_custom_all_reduce.py`: 2 passed.
- Ruff check and format: passed.
- GPU microbenchmark: pending an exclusive TP8 window.

The runtime route is explicit and default-off via `VLLM_SM70_TP8_NONFULL_CUSTOM_AR=1`.


yangzhuxinyzx:
Closed after backlog audit. The TP8 GPU microbenchmark and endpoint quality gate were never run, while later branches already screened and rejected related collective variants. A default-off runtime route without its stated hardware gate is not mergeable.

PR #168, opened 2026-08-02
[Kernel] Add SM70 DeepSeek V4 FP16 GEMV

yangzhuxinyzx:
## Purpose

Add a fixed-shape FP16-to-FP32 GEMV route for the remaining DeepSeek-V4-Flash router and compressor decode cost on V100. The route removes cuBLAS split-K reducer overhead for five exact M=1 shapes while preserving FP32 accumulation semantics.

Base SHA: `2872c6a3af4897fc59855b8a59310fa3789ab5fc`

## Implementation

- Dispatch one SM70 program per output row with FP16 inputs and FP32 accumulation.
- Cover router, indexer-weight, C4 indexer compressor, C4 main compressor, and C128 main compressor shapes.
- Preserve all generic and non-SM70 paths.
- Keep production dispatch explicit and default-off behind `VLLM_SM70_DSV4_FP16_GEMV=1`.
- Retain the rejected fused-router prototype as benchmark evidence only.

## Test Plan And Result

- Real checkpoint CUDA Graph microbenchmark: all five accepted shapes improve by 1.18x to 1.79x.
- Summed operator projection: 0.463 ms/token; joined C4 multi-stream envelope: 42.004 to 31.908 us.
- Sixteen real-weight router seeds preserve top-6 IDs; maximum normalized routed-weight error is 3.61e-5.
- Compressor maximum absolute error is at most 2.39e-6.
- V100 focused graph/dispatch tests: all five shapes plus the default-off gate pass.
- Exact combined 8xV100 integration tree in PR #180: full SM70 extension link passes; FP16 GEMV, mHC, and route suite contributes to 24 passing focused tests; official-sampling text-health suite passes.

## Rejected Path

The fused router candidate saves only 0.134 ms/token over 40 real layers, below the 0.2 ms/token admission threshold, so it is not wired into production dispatch.

## Evidence And Risk

Detailed measurements and retained artifact paths are in `docs/design/sm70_deepseek_v4_fp16_gemv.md`. The accepted route remains default-off pending long-context, concurrency, and second-host promotion gates.


yangzhuxinyzx:
Final promotion audit: PR title/body were corrected to reflect the accepted production dispatch rather than the early benchmark-only state. The branch is synchronized with current onecat/main; its diff now contains only the five-shape FP16 GEMV route, tests, benchmarks, and design record. Exact combined 8xV100 build/route/quality gates in #180 pass. The route remains default-off pending long-context, concurrency, and second-host promotion.

PR #164, opened 2026-08-02
[Kernel] Optimize SM70 MXFP4 grouped decode

yangzhuxinyzx:
## Purpose

Reduce DeepSeek-V4-Flash TP8 batch-one MXFP4 MoE decode latency after the
active-expert route in PR #160. Six one-row experts now execute as one
TurboMind grouped W13 launch and one grouped W2 launch. An exact direct top-6
route also removes the remaining generic 256-expert routing setup.

Base SHA: `2872c6a3af4897fc59855b8a59310fa3789ab5fc`

## Test Plan

- Exact-shape CUDA Graph microbenchmarks with dynamic route replay
- Bitwise checks for routing metadata, W13, clamp-10 SwiGLU, W2, and output
- Four-stream attention projection benchmark to reject false single-stream wins
- TP8 exact 1024-input/256-output endpoint A/B on the PR #160 + PR #163 stack
- Fresh graph-node trace and official-sampling quality gate

## Test Result

- Compact W13: 0.1156 -> 0.0286 ms, bitwise equal
- Compact W2: 0.0433 -> 0.0117 ms, bitwise equal
- Graph trace: MXFP4 7.542 -> 1.957 ms/token and 516 -> 86 launches/token
- Temporary endpoint candidate: 33.853 -> 28.961 ms/token, about 34.53 tok/s
- Exact direct top-6: 0.101144 -> 0.051226 ms/layer across five seeds
- Direct route projected saving: 2.149 ms/token across 43 layers
- Direct route remains bitwise equal after CUDA Graph route-ID mutation
- `pytest -q tests/quantization/test_sm70_mxfp4_moe.py`: 19 passed
- Python Ruff, format, py_compile, and `git diff --check`: passed
- Full SM70 native extension build and operator microbenchmarks: passed

The 28.961 ms endpoint included an experimental FP8 WQA/WKV split. A
four-stream benchmark proved that split regresses the real overlapped schedule
by about 0.556 ms/token, so all split implementation code was removed. The
reported endpoint therefore proves the compact MXFP4 route transfers to the
model, but is not the final accumulated result.

Rejected marginal paths are retained as benchmarks:

- generic grouped dispatch and CTA_N-only tuning regress;
- FP8 WQA/WKV split is a false single-stream win;
- mHC `tile_n=1` projects only about 0.035 ms/token.

Detailed evidence and artifact paths are in
`docs/design/sm70_deepseek_v4_mxfp4_grouped_decode.md`.

## Remaining Gates

[...]

PR #169, opened 2026-08-02
[Benchmark] Correct SM70 FP8 dense hotspot timing

yangzhuxinyzx:
## Purpose

Correct the DeepSeek-V4-Flash SM70 FP8 CUDA Graph microbenchmark and record the rejected exact CTA_N=64/32 screen. No unproven kernel is enabled.

Base/dependency: PR #164.

## Test Plan

- Warm the TurboMind workspace on the capture stream before graph capture
- Delimit a single steady replay for NSYS/NCU
- Hold split-K=5 constant when comparing CTA_N=128/64/32
- Keep direct-vs-graph output parity

## Test Result

- Corrected six-shape standalone projection: 3.906 ms/token
- CTA_N=128: 21.212 us
- CTA_N=64: 22.558 us (+6.3%)
- CTA_N=32: 23.541 us (+11.0%)
- Candidate kernels rejected before endpoint testing
- Ruff check/format and git diff --check pass

### DIFF
--- a/benchmarks/benchmark_sm70_deepseek_v4_fp8_dense.py
+++ b/benchmarks/benchmark_sm70_deepseek_v4_fp8_dense.py
@@ -8,6 +8,7 @@ import argparse
 import hashlib
 import json
 import statistics
+from collections.abc import Callable
 from dataclasses import asdict, dataclass
 from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


yangzhuxinyzx:
Closed as an archival benchmark branch. It records corrected timing and rejected CTA_N candidates on top of the already integrated dependency branch; no accepted production kernel remains to merge.

PR #170, opened 2026-08-02
[Core] Reduce SM70 TP8 CUDA Graph collective bubbles

yangzhuxinyzx:
## Purpose

Remove the CUDA Graph/NCCL launch bubble before DeepSeek-V4-Flash's exact TP8 8 KiB FP16 all-reduces on the two-clique V100 topology.

Base/dependency: PR #164.

## Implementation

- Validate two four-rank P2P cliques plus paired cross-clique links.
- Open only required IPC mappings.
- Form fixed-rank-order FP32 clique partials, exchange over the paired NV2 link, and downcast once to FP16.
- Alternate two graph-safe signal/partial slots and acknowledge clique input consumption concurrently with the cross-clique pair exchange.
- Dispatch only for SM70, TP8, FP16, 4096 elements, non-fully-connected topology, and an explicit default-off flag.
- Preserve all existing communicator paths outside that exact contract.

## Test Result

- Exact graph-join micro: 39.384 us NCCL -> 32.154 us custom per call; 0.629 ms projected over 87 calls.
- Pure graph stress: 8,700 collectives complete at 17.099 us/call.
- Component output: all ranks bitwise equal; max/mean error vs FP32 is 0.061523/0.006093, versus NCCL 0.136719/0.014117.
- Matching all-reduce-only 1K/256 endpoint, three seeds: 25.630149 -> 23.460869 ms/token (-2.169281 ms, -8.46%); 39.0166 -> 42.6242 token/s (+9.25%).
- Nsight critical path: all-input-ready to collective completion 63.740 -> 40.835 us; 1.993 ms/token projected critical saving.
- Stacked QK-D-split endpoint after skew fix: 20.768 / 20.758 / 20.770 ms/token, mean 20.765 ms, all three requests complete 256/256 tokens. This is combination evidence, not an all-reduce-only A/B.
- Targeted pytest: 10 passed.
- Remote SM70 `_C` rebuild and git diff checks pass.

## Rejected Paths

- NCCL Tree improved isolated transport but moved the endpoint only 0.014 ms.
- 256-thread multi-item CTA deadlocked in synchronization; removed. The accepted kernel remains 512 threads.
- One signal/partial slot stopped after 22 endpoint tokens under larger graph skew.
- Partial-only double buffering still entered synchronization spin.
- A serialized completion barrier was stable but 0.290 ms/token slower than the accepted concurrent acknowledgement.

## Remaining Gate

The route remains default-off pending the model-specific official-sampling quality suite and validation on a second matching topology.


yangzhuxinyzx:
Final promotion audit: synchronized with the #164 merge on onecat/main. An exact full SM70 extension rebuild exposed a multiple-definition link failure because the header-defined TP8 kernel had external linkage; commit d7bf0c102b internalizes that symbol, and the complete _C link now passes. All 10 focused TP8 collective tests pass, the exact combined 8xV100 route/quality gate in #180 passes, and the PR diff now contains only the hierarchical all-reduce scope. The route remains default-off pending second-topology, long-context, and concurrency gates.

PR #163, opened 2026-08-02
[Kernel] Parallelize SM70 DeepSeek V4 sparse MLA

yangzhuxinyzx:
## Purpose

Replace the single-CTA SM70 DeepSeek V4 sparse MLA decode bottleneck with a
graph-safe FP32 split-K attention-state reduction. SWA-only, C4, and C128
remain independently opt-in until the remaining quality and long-context gates
pass.

Base SHA: `dd462e37f2552f3e038f1ed7128e62bd7b4ab0d7`

## Test Plan

- Exact-shape SWA/C4/C128 CUDA Graph microbenchmarks on V100
- Full 256-code E4M3 decode comparison
- q=1/q=2 graph replay and multi-length numerical comparison
- TP8 exact 1024-input/256-output endpoint benchmark
- Nsight Systems TP8 CUDA Graph node trace
- Accumulated SWA endpoint attribution (pending)
- Deterministic logits/model-specific semantic quality (pending)
- Long-context endpoint sweep (pending)

## Test Result

- C4 main128+extra320: 1.957 ms -> 0.101 ms; max abs error <= 1.53e-5
- C128 main128+extra10: 0.581 ms -> 0.078 ms; max abs error <= 3.05e-5
- SWA-only main128, three seeds: 0.586 -> 0.059 ms/layer; about 9.9x; max abs
  error <= 3.05e-5
- SWA q=2 CUDA Graph replay is finite and remains within the same error bound
- C128 extra 512/1024/2048: 0.100/0.119/0.152 ms; max abs error <= 7.63e-6
- E4M3: all 254 finite values bitwise equal; NaN masks equal
- `pytest -q tests/models/test_deepseek_v4_sm70_routes.py`: 9 passed
- `pytest -q tests/quantization/test_sm70_mxfp4_moe.py`: 13 passed
- Unprofiled C4+C128 TP8 TPOT: 76.268 -> 33.853 ms (-55.61%)
- Unprofiled C4+C128 decode throughput: 13.112 -> 29.539 tok/s (+125.29%)
- Sparse MLA GPU service: 46.920 -> 4.392 ms/token (-90.64%)
- Node trace: 255 x 8 replay ranges, 247 steady steps, 98.20% coverage

The 33.853 ms endpoint predates SWA split-K. Its trace still contains two
unsplit SWA calls totaling 1.094 ms/token; the new microbenchmark projects
about 1.05 ms/token additional service reduction, but this remains a
projection until the accumulated endpoint gate completes.

[...]

PR #171, opened 2026-08-02
[Kernel] Reduce SM70 sparse MLA register stalls

yangzhuxinyzx:
## Purpose

Reduce the remaining DeepSeek-V4-Flash SM70 sparse MLA decode latency after PR #163 by attacking its measured 255-register, low-occupancy QK bottleneck.

Base/dependency: PR #163. The stacked endpoint also requires the graph-skew-safe revision of PR #170 when hierarchical TP8 all-reduce is enabled.

## Implementation

- Split the 512-dimensional QK dot into eight 64-dimensional tiles.
- Increase C4 QK work from 40 to 320 CTAs.
- Reduce each score in fixed D-tile order, apply the existing scale/softmax, then run 64-dimensional PV tiles.
- Allocate FP32 QK/state/accumulator and FP16 probability scratch through the graph-safe workspace manager.
- Keep the route exact-shape, SM70, split-K, and explicit default-off.

## Test Result

- CUDA Graph micro, C4: 0.08565 -> 0.03061 ms (-64.3%), max abs 7.63e-6.
- C128: 0.06286 -> 0.02269 ms (-63.9%), max abs 3.05e-5.
- SWA-only: 0.06278 -> 0.02153 ms (-65.7%), max abs 3.05e-5.
- Direct old/candidate output: C4 bitwise exact; C128 and SWA differ in one FP16 element with max abs 3.8147e-6 and 1.9073e-6.
- C4 NCU: 255 -> 95 registers/thread, 33.28 -> 4.10 KiB dynamic shared memory, 6.25% -> 21.65% achieved occupancy, 40 -> 320 CTAs.
- Compute Sanitizer C4: zero errors.
- Latest graph-node sparse service: about 3.48 -> 1.35 ms/token (-2.13 ms).
- Stacked 1K/256 endpoint: 23.460869 -> 20.765477 ms/token (-11.49%); 42.6242 -> 48.1569 token/s (+12.98%). Candidate samples are 20.767631 / 20.758487 / 20.770312 ms and all complete 256/256 tokens.
- Ruff check/format, focused pytest (2 passed), and git diff checks pass.

## Rejected Paths

- Score-only split then PV: 0.10887 ms C4.
- QK D=32: NaN output.
- QK D=128: 0.03716 ms, slower than D=64.
- BLOCK_H=4: 0.04340-0.04503 ms.
- Two/eight QK warps: 0.03635/0.03743 ms.

## Remaining Gate

Keep default-off pending the independent official-sampling quality suite, long-context/concurrent decode sweep, and second-host validation.


yangzhuxinyzx:
Final promotion audit: synchronized with the #163 merge on onecat/main. The PR now contains only the QK D-split delta. The resulting sparse stack is byte-identical to the exact 8xV100 integration runtime in #180; focused route tests and the combined official-sampling quality gate passed. This route remains explicit default-off pending long-context, concurrency, and second-host validation.

PR #173, opened 2026-08-03
[Benchmark] Analyze SM70 TP8 collective skew

yangzhuxinyzx:
## Purpose

Analyze the remaining DeepSeek-V4-Flash TP8 collective critical path on 8x V100. Screen and reject a two-CTA hierarchical all-reduce, then align CUDA Graph collective nodes by ordinal to localize rank-arrival skew and shared/routed MoE scheduling.

Dependencies: #169, #170, #171. Base SHA: `3ec0c68c6596d6ab31fbdee9fa676254a52c2b7d`.

## Test Plan

- Exact 4096-element FP16 TP8 CUDA Graph stress and joined-work microbenchmarks.
- Offline analysis of the latest 63-replay graph-node sqlite with fixed 87-call ordinal alignment.
- Exact-shape single-GPU and TP8 shared-MoE scheduling graphs.
- Require bitwise equality on initial and changed-input graph replays.

## Test Result

- Two CTA: 18.710 -> 19.879 us/call pure and 32.498 -> 35.035 us/call joined; rejected and removed.
- Steady collective arrival skew: 31.578 us/call, 2.747 ms/token summed. MoE segment skew: 41.363 us/layer.
- Router-first shared stream: 81.741 -> 75.615 us/layer single GPU, but only 102.664 -> 101.142 us/layer in the TP8 join, 0.065 ms/token projected; rejected before endpoint.
- Initial and dynamic TP8 outputs are bitwise equal on all ranks.
- Ruff, format, Python compile, and diff checks pass.

Raw artifacts are recorded in `docs/design/sm70_deepseek_v4_tp8_two_cta_allreduce.md`. No rejected production gate or kernel remains in the final source diff.


yangzhuxinyzx:
Closed as completed analysis, not a production change. The two-CTA and router-first candidates were explicitly rejected and the branch says no rejected kernel remains. Its useful outcome is the recorded diagnosis, not a merge into current main.

PR #175, opened 2026-08-03
[Kernel] Stage SM70 DeepSeek V4 mHC decode

yangzhuxinyzx:
## Purpose

Remove repeated post-mapping work from the DeepSeek V4 Flash M=1 mHC decode path on V100 while preserving FP32 accumulation order and operator outputs.

Dependency: PR #168.
Current head: `3441f09c10ded37c5a9852df144e17b512f4972f`

## Implementation

- Add an SM70-only FP32 post staging kernel and a dot kernel that retain split-K=8, tile_n=2, and the original reduction order.
- Gate to FP16, M=1, H=4096, hc=4, and SM70; preserve every other mHC shape.
- Enable the accepted route by default with `VLLM_SM70_DSV4_MHC_FP32_STAGE=1`; `0` restores the original fused kernel.
- Add exact-shape, all-85-call, OpenAI streaming, and focused bitwise tests.

## Test Result

- One-layer CUDA Graph microbenchmark: 27.881 to 14.668 us/call, bitwise exact.
- All 85 real checkpoint calls: 2.84065 to 1.52876 ms/token; all four outputs bitwise equal for every call.
- V100 focused mHC GPU test passes; the exact combined integration suite in #180 includes 24 passing FP16 GEMV, mHC, and DeepSeek V4 route tests.
- Exact integration-tree TP8 A/B, input 1024/output 256, FP8 MLA KV, official temperature=1.0/top_p=1.0 sampling:
  - route off: median 20.4009 ms/token, 49.02 token/s;
  - route on: median 19.4566 ms/token, 51.40 token/s;
  - TPOT -4.63%, decode throughput +4.85%.
- Official-sampling chat text-health suite passes, including a natural-stop request. Independent TP8 services are not a deterministic token oracle under the existing multi-stream custom-all-reduce stack, so numerical acceptance uses the bitwise all-layer operator oracle plus text health.

## Evidence And Risk

Detailed measurements and retained artifacts are recorded in `docs/design/sm70_deepseek_v4_mhc_decode.md` and PR #180. The route adds about 5.3 MiB of CUDA Graph pool staging per rank and has an environment rollback.


yangzhuxinyzx:
Final promotion audit: synchronized with the #168 merge on current onecat/main. The PR diff is limited to mHC FP32 staging, its tests, benchmarks, and design record. Exact all-layer bitwise oracles, full combined SM70 link, route logs, official-sampling quality, and matched 8xV100 endpoint A/B all pass as recorded in #180.

PR #176, opened 2026-08-03
[Benchmark] Reject SM70 TP8 partitioned all-reduce

yangzhuxinyzx:
## Purpose
Record and retire a conflict-free partitioned TP8 hierarchical all-reduce experiment for the fixed 4096-element SM70 decode collective. Candidate code is not retained.

## Test Plan
- TP8 V100 CUDA Graph pure-collective matched A/B
- TP8 V100 multi-stream graph-join matched A/B
- FP32 reference error and bitwise cross-rank/output hash checks

## Test Result
- Accepted: 19.849 us pure, 37.898 us graph-join
- CTA gather: 24.696 us pure, 43.763 us graph-join
- Warp gather: 26.463 us pure, 45.541 us graph-join
- Both candidates matched accepted output hashes and were bitwise equal across all eight ranks
- Rejected because graph-join regressed by 5.865-7.643 us/call

### DIFF
--- a/docs/design/sm70_deepseek_v4_tp8_graph_allreduce.md
+++ b/docs/design/sm70_deepseek_v4_tp8_graph_allreduce.md
@@ -165,6 +165,26 @@ the 0.2 ms/token continuation threshold, so the deferred-poll code was removed.
 The remaining TP opportunity is upstream rank-arrival skew or a larger
 compute/collective boundary change, not further flag-loop tuning.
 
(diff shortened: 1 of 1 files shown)


yangzhuxinyzx:
Closed because this PR explicitly records a rejected partitioned all-reduce experiment and retains no accepted candidate code. It is an archival result, not a mainline change.

PR #177, opened 2026-08-03
[Benchmark] Reject SM70 TP8 arrival prefetch

yangzhuxinyzx:
## Purpose
Record and retire an exact SM70 TP8 arrival-time peer-prefetch experiment. Candidate code is not retained.

## Test Plan
- TP8 V100 CUDA Graph pure-collective smoke
- TP8 V100 matched multi-stream graph-join A/B
- FP32 reference error, output SHA, and cross-rank bitwise checks

## Test Result
- Accepted graph-join: 37.919 us/call
- Arrival prefetch graph-join: 42.474 us/call
- Projected regression: 0.396 ms/token over 87 calls
- Candidate matched accepted output SHA and was bitwise equal across all eight ranks
- Rejected because polling/register lifetime exceeds the load work hidden under rank skew


yangzhuxinyzx:
Closed because this PR explicitly records a rejected arrival-prefetch experiment with a projected regression. It is an archival result, not a mainline change.

PR #178, opened 2026-08-03
[Kernel] Stage SM70 compressor state-save fusion

yangzhuxinyzx:
## Purpose

Stage an SM70 batch-one DeepSeek V4 compressor state-save fusion without changing the default runtime path.

Base SHA: `c6c72ed0f4ff420d03312b5bc1b5ad4712e9d78f`
Head SHA: `d19f9b98da`

## Implementation

- Fold current KV/score/APE state write into the existing Triton compressor kernel.
- Gate on SM70, M=1, FULL CUDA Graph, FP8 cache, and `VLLM_SM70_DSV4_FUSED_COMPRESSOR_SAVE=1`.
- Keep MXFP4 cache and every unsupported shape on the existing path.
- Add an exact-shape CUDA Graph microbenchmark and update AOT/signature coverage.

## Test Plan And Result

- `py_compile`, Ruff, Ruff format, and `git diff --check`: pass.
- V100 CUDA Graph microbenchmark, 6 C4/C128 boundary/non-boundary cases, 5 x 1000 replays: state and KV cache bitwise exact.
- Weighted microbenchmark projection: `0.020449 ms/token` saved. This is not an endpoint result and is too small for acceptance.
- Targeted FP8 Triton tests: 6 passed.
- Existing MXFP4 direct Triton cases: 6 fail because Triton emits unsupported SM70 `cvt.e2m1x2`; this candidate never enables fusion for MXFP4 cache.

Artifact: `/home/fudanwl/v100-worktrees/runs/dsv4-compressor-save-fusion-micro-20260803/result.json`

## Status

Draft only. Default remains disabled. Promotion requires matched 1K/256 endpoint A/B and output-quality validation; current microbenchmark evidence does not justify enabling it.


yangzhuxinyzx:
Closed after acceptance audit. The route saves only a projected 0.020449 ms per token, remains default-off, lacks matched endpoint and output-quality evidence, and had unsupported MXFP4 test failures. It does not meet the merge threshold.

PR #179, opened 2026-08-03
[Kernel] Stage SM70 MXFP4 grouped prefill

yangzhuxinyzx:
## Purpose

Preserve and review the in-progress SM70 DeepSeek V4 MXFP4 grouped-prefill route without adding it to the accepted active-expert PR.

Base SHA: `2872c6a3af4897fc59855b8a59310fa3789ab5fc`
Head SHA: `ea72ec086b`

## Implementation

- Route populated prefill experts through the existing TurboMind grouped MoE entry point.
- Keep the candidate opt-in behind `VLLM_SM70_MXFP4_MOE_GROUPED_PREFILL=1`.
- Strengthen the dynamic-routing CUDA Graph benchmark.
- Add operator-prefill and timed streaming endpoint benchmark drivers.

## Test Plan And Result

- `py_compile`, Ruff, Ruff format, and `git diff --check`: pass.
- GPU rebuild, numerical exactness, prefill A/B, and endpoint quality/performance: not run yet.

## Status

WIP Draft only. The default route is unchanged. Do not promote or enable until the TP8 V100 exactness, repeatability, matched prefill, and decode-neutrality gates pass.

### DIFF
--- a/benchmarks/benchmark_sm70_mxfp4_moe_active_experts.py
+++ b/benchmarks/benchmark_sm70_mxfp4_moe_active_experts.py
@@ -149,29 +149,49 @@ def _validate_permute_contract(
         top_k, num_experts
     )
     workspace = torch.empty(workspace_size, dtype=torch.int8, device=device)
-
-    torch.ops._moe_C.moe_permute_with_scratch(
-        x,
-        topk_ids,
-        token_expert_indices,
-        None,
-        num_experts,
-        num_experts,
-        top_k,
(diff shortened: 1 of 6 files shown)


yangzhuxinyzx:
Closed as unvalidated WIP. The PR itself records that GPU rebuild, numerical exactness, prefill A/B, endpoint quality, and decode neutrality were never run. No source route should be retained before those gates pass.

PR #181, opened 2026-08-03
[Core] Optimize DeepSeek V4 DSpark N7 on SM70

yangzhuxinyzx:
## Purpose

Optimize DeepSeek V4 Flash DSpark N=7 on SM70 while correcting the matched no-speculation baseline. The PR removes the false 4K short-context penalty, fixes V100 FP8 Index-K prefill, and replaces the verifier's per-expert MXFP4 launch explosion with an exact slot-grouped TurboMind path.

Base SHA: `48e89751b4b98c18e1be6506dca15f015155d068`
Current head: `61fc4e763e0e261a9b3882525d17348aedbfec44`

## Implementation

- Capture and dispatch a bounded no-spec B1 graph from the real C4 Indexer threshold; safely fall back above context 2048.
- Decode contiguous FP8 Index-K storage through the existing SM70 software E4M3 path as `uint8`.
- Allocate graph-stable MXFP4 MoE buffers through M=8.
- First compact dense-256 verifier work to a fixed active-48 table entirely on GPU.
- For exact M=8, retain all 48 routed slots as independent one-row groups and dispatch each W13/W2 stage in one TurboMind launch. Repeated experts reuse their packed weight pointer; arithmetic and accumulation order are unchanged.
- Keep a separate immutable slot-offset buffer so M=2-M=7 graph replay cannot pollute M=8 metadata.
- Preserve the old active-48 loop as a default-off rollback. Enable the accepted candidate with `VLLM_SM70_MXFP4_MOE_GROUPED_M8=1`.
- Extend synchronized SM70 MTP profiling to DSpark/DFlash and add exact M=8 route-distribution benchmarks.

## Performance

Contract: TP8 on 8x V100-SXM2-32GB, DeepSeek-V4-Flash FP8 dense + MXFP4 experts, `fp8_ds_mla`, exact 1024 input / 256 output, official temperature=1.0/top_p=1.0, CUDA Graph, no eager.

| Result | Original | Active-48 | Slot-grouped | Final vs active-48 |
|---|---:|---:|---:|---:|
| No-MTP TPOT | 19.342 ms false 4K graph | 15.357 ms bounded graph | unchanged | n/a |
| DSpark N7 median TPOT | 82.007 ms | 60.441 ms | 32.873 ms | -45.6% |
| DSpark N7 median throughput | 12.194 token/s | 16.545 token/s | 30.420 token/s | +83.9% |
| M=8 target forward | 126.531 ms | 78.779 ms | 42.394 ms | -46.2% |
| Target + draft GPU subtotal | 137.444 ms | 86.416 ms | 47.923 ms | -44.5% |

Matched slot-grouped seeds produce 29.747, 32.416, and 30.420 token/s. Aggregate emitted tokens per stream chunk are 1.580 for active-48 and 1.571 for slot-grouped, so acceptance does not explain the gain.

Exact one-layer CUDA Graph microbenchmarks cover permute, W13, clamped SwiGLU, W2, and unpermute:

| Route distribution | Active-48 loop | Slot-grouped | Speedup |
|---|---:|---:|---:|
| Mixed, 34 unique experts | 1.535 ms | 0.472 ms | 3.25x |
| 48 unique experts | 1.380 ms | 0.184 ms | 7.49x |
| Six experts with eight rows each | 0.708 ms | 0.281 ms | 2.52x |

All intermediate stages and changed-route graph replays are bitwise equal, with max absolute error 0. The generic multi-row grouped scheduler was explicitly rejected after seed-dependent W13 drift and hot-expert W2 drift up to 0.25.

## Test Plan And Results

- All selected pre-commit hooks: pass, including Ruff, clang-format, mypy, markdownlint, SPDX, and the torch CUDA API policy.
- Local MXFP4 MoE suite: `21 passed, 1 skipped` (GPU test skipped locally).
- Remote V100 MXFP4 MoE plus environment getter tests: `24 passed`.
- Three M=8 route-distribution CUDA Graph microbenchmarks: bitwise at W13, SwiGLU, W2, final output, and dynamic replay.
- Full TP8 graph capture and three matched 1024/256 seeds: pass.
- Synchronized profile: 11 steady eight-step intervals, target forward median 42.394 ms and GPU subtotal 47.923 ms.
- Official-sampling natural-stop quality gate: coherent two-sentence response, 66 completion tokens, `finish_reason=stop`.
- Forced context-2048 transition and exact 2304-token FP8 prefill gates from the prior commit: pass.
- Same-contract no-MTP regression from the prior commit: 15.377 ms/token, within 0.13% noise of the accepted 15.357 ms baseline.

Raw commands, artifacts, rejected paths, and the remaining round budget are recorded in `docs/design/sm70_deepseek_v4_dspark_n7_latest_main.md`.

## Known CI Baseline

The GitHub `pre-commit` job currently fails because the workflow runs `--all-files --hook-stage manual` over inherited repository debt, not because of the six files in this commit. The current log reformats 72 unrelated Python files and 120 unrelated C/CUDA files, then reports existing vendored TurboMind/Flash-V100 typos plus repository-wide mypy, SPDX, forbidden-import, and torch-CUDA findings. The same branch check already failed at heads `3765c56a96` and `4447bfe7da` before the slot-grouped commit. All hooks restricted to this PR's changed files pass locally. This should be resolved in a separate baseline-lint PR rather than mixing hundreds of unrelated formatting edits into this kernel PR.


valentijnvenus:
Nice one!

PR #184, opened 2026-08-03
[Core] Complete DeepSeek V4 8K long-context path

yangzhuxinyzx:
## Purpose

Complete the DeepSeek V4 8K-prefill long-context path on SM70/V100 while preserving the existing `max_num_batched_tokens=8192` deployment contract.

Task base: `3993136386e5088fba5ef27b68cfb4b5b1f57621`
Current head: `b44f44ea92`

## Changes

- Replace paged transient compressor state with a bounded private SM70 state ring for the supported single-request, no-prefix-cache path; preserve the paged path everywhere else.
- Stage the ratio-128 private history through workspace memory and add byte-exact compressor-state regression coverage.
- Use compressed KV length, rather than original token length, for SM70 indexer logits and top-k workspace sizing.
- Enable the automatic DeepSeek V4 short-context CUDA-graph bucket for MTP and retain explicit MTP overrides.
- Align SparseSWA's DSpark parallel-drafting decode threshold with Indexer/FlashMLA: `1 + 2 * N` (15 for N=7).

## Root Cause

Long-context MTP used the original sequence length for compressed-indexer workspace sizing, inflating the SM70 path. In parallel DSpark drafting, SparseSWA classified an N=7 decode batch with an 8-token threshold instead of 15, causing backend split disagreement. The compressor also retained a much larger transient paged state than the supported single-request route needs.

## Validation

- `pytest -q tests/v1/attention/test_attention_splitting.py -k test_deepseek_sparse_swa_spec_decode_threshold`: 2 passed.
- V100 deployment environment: `pytest -q tests/v1/cudagraph/test_cudagraph_dispatch.py -k test_dsv4_context_bucket_is_derived_for_mtp_on_sm70`: 1 passed.
- `ruff check` and `ruff format --check` on all 15 changed Python files: passed.
- `python -m compileall` on changed source modules: passed.
- `git diff --check`: passed.
- The deployed V100 source hashes match the submitted SparseSWA, indexer, and CUDA-graph dispatcher files.

## Runtime Evidence

TP8 / DSpark N=7 / FP8-KV / CUDA Graph startup passed with `max_model_len=307200`; long code continuation was text-healthy. Artifacts: `/home/fudanwl/v100-worktrees/runs/dsv4-8k-longctx-20260804`.

## Remaining Gate

The new GPU-only compressor-ring and SM70 indexer kernel tests were not rerun on the active eight-GPU API host, to avoid contending with the live service. Run them on an idle V100 allocation before merging this Draft PR.


yangzhuxinyzx:
Closed as incomplete and conflicting. The GPU-only compressor-ring and indexer tests named as required gates were never run, and the evidence is a startup/text-health smoke rather than matched output-quality acceptance. Re-port the minimal fixes to current main only after those exact hardware gates are available.

PR #200, opened 2026-08-13
[Kernel][SM70] MLA prefill: selectable Flash-V100 route for uniform 256/256 head dims

kkobold:
## Purpose

Make FlashAttention MLA prefill **selectable on SM 7.0** so models with uniform 256/256 head dims (e.g. GLM-MoE-Lite) can pick FLASH_ATTN prefill instead of failing at engine init with:

```
No valid MLA prefill backend found ... {FLASH_ATTN: [required dependencies not available]}
```

Before this PR, the FlashAttention MLA prefill selector gates itself off on Volta, and there is no fallback that can serve MLA prefill there. This PR wires the SM70 route (via the in-tree Flash-V100 dense-LSE op) into the selector, and keeps non-selectable configs REJECTED at load time — so the clean-error guarantee for anything the SM70 route cannot serve is preserved.

Three edits close the selectability gap:

1. `is_available()` returns True on SM 7.0 when the in-tree Flash-V100 dense-LSE op is usable.
2. `validate_configuration()` override rejects any config the SM70 route cannot run: non-fp16 dtype, non-uniform head dims (e.g. DeepSeek 192/128 asymmetric qk/v), or a non-compiled head-dim tile.
3. `__init__` assert relaxed on the SM70 route.

Load-time rejection preserves clean-error behavior: a config the backend can't serve is refused up front, never selected-then-crashed.

Files: `vllm/v1/attention/backends/flash_attn_v100.py`, `vllm/v1/attention/backends/mla/prefill/flash_attn.py`, `vllm/v1/attention/backends/mla/prefill/selector.py`. 3 files, +425 / −5 lines.

**Deferred (not in this PR):** the DeepSeek 192/128 asymmetric slice is provably held out by the selector at load time. A follow-up PR covers that.

## Test Plan

Hardware: SM 7.0 (Tesla V100 SXM2 32GB), TP4.

1. **Op-level numerics** — three-arm merge-corner gate on the SM70 route:
   - Unmasked driver (negative control) — expect NaN poisoning
   - Fixed masked driver — expect attention reconstruction within fp16 noise
   - Random negative control — expect large error
2. **Engine selectability** — load a model with uniform 256/256 MLA head dims (e.g. `rt-base-w4g128`, GLM-MoE-Lite lineage) on SM70 TP4 and confirm the selector picks `FLASH_ATTN` MLA prefill without crashing.
3. **End-to-end coherence** — greedy answers on canonical prompts (capital of France, arithmetic, a short story) — verify correctness, not just non-crash.
4. **Regression suite** — a delivered A/B against the prior release version on the same served config, covering decode tok/s, prefill prompt tok/s, TPOT, and HumanEval mean.

## Test Result

**Op-level numerics (real SM70, three-arm merge-corner gate):**

| Arm | Result |
|---|---|
| Unmasked driver | NaN-frac 0.667 (2/3 requests poisoned) — REJECTED |
| Fixed masked driver | Attention reconstructed at `rel_rms 2.42e-4` vs fp32 oracle |
| Random negative control | rel_rms 0.41 — REJECTED |

**Live engine boot (rt-base-w4g128 on V100 TP4):**

```
selector.py:191 Using FLASH_ATTN MLA prefill backend
Using TRITON_MLA attention backend
```

Greedy answers on the loaded engine:
- "The capital of France is Paris… London" — correct
- "17 plus 26 equals 43" — correct
- Fluent short-story generation

Before this PR: engine init raised at the selector, no MLA prefill backend available on SM 7.0.

**Delivered A/B (against prior release version, same served config):**

| Metric | Delta | Interpretation |
|---|---|---|
| Decode tok/s | −1.8…+0.1% | inside spread |
| Prefill prompt tok/s | −1.0…+0.2% | inside spread |
| TPOT | +0.5…+1.1% | inside spread |
| HumanEval mean | inside own spread | no regression |

Flat is the predicted result: against the deployed served config (AWQ, fp8_e5m2 KV, no speculative decoding) this patch set barely touches the served hot path. The suite's job was to prove no regression, which it does. The AVAILABILITY win — MLA prefill from "unselectable / crashes" to "engine-selectable and correct" — is the value here.

---

<details>
<summary>Essential Elements of an Effective PR Description Checklist</summary>

- [x] The purpose of the PR
- [x] The test plan
- [x] The test results
- [ ] (Optional) Documentation update — happy to add supported-models note if reviewers want one
</details>

### DIFF
--- a/vllm/v1/attention/backends/flash_attn_v100.py
+++ b/vllm/v1/attention/backends/flash_attn_v100.py
@@ -854,6 +854,195 @@ def flash_v100_dense_prefill(
     return output
 
 
+# 0014: LSE-emitting dense varlen prefill (revives archived 0043 + discharges 0044)
+# ----------------------------------------------------------------------------
+# The SM70 kernel ALWAYS computes the softmax log-sum-exp
+# (flash-attention-v100/kernel/fused_mha_forward.cu:688-692 writes rowmax+log(rowsum),
+# :1149 returns it as outputs[1]). What flash_v100_dense_prefill drops is that LSE,
+# which the MLA context-chunk / prefix-cache merge needs (mla_attention.py:2109-2136,
+# :2305-2312). This op exposes it and additionally accepts an INDEPENDENT cu_seqlens_k,
+# so the cross-attention shape of run_prefill_context_chunk (M new-token queries vs N
+# gathered context keys, M != N, unmasked) is expressible.
+_flash_attn_forward_lse = None
+_flash_attn_forward_lse_checked = False
+
+
+def _get_flash_dense_forward():
+    """Lazy-load the private LSE-capable forward entry of the FA-V100 wheel."""
+    global _flash_attn_forward_lse, _flash_attn_forward_lse_checked
+    if not _flash_attn_forward_lse_checked:
+        _flash_attn_forward_lse_checked = True
+        try:
+            from flash_attn_v100.flash_attn_interface import _flash_attn_forward
+
+            _flash_attn_forward_lse = _flash_attn_forward
+        except (ImportError, AttributeError):
+            _flash_attn_forward_lse = None
+    return _flash_attn_forward_lse
+
+
+def flash_v100_dense_prefill_lse_available() -> bool:
+    return _get_flash_dense_forward() is not None
+
+
+def flash_v100_dense_prefill_lse(
+    query: torch.Tensor,
+    key: torch.Tensor,
+    value: torch.Tensor,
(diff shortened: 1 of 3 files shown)


yangzhuxinyzx:
Merged through audited PR #223. The two-parent audit merge preserves this commit and authorship while fixing the real-V100 availability-stub and causal-offset-identity blockers, and adds load-time gates, metadata validation, empty-K neutral handling, FP64 numerical checks, and full policy/layout regression coverage.

PR #201, opened 2026-08-13
[Kernel][GDN] Chunk-kernel improvements bundle: solve_tril, chunk-indices, chunk_o

kkobold:
## Purpose

Three orthogonal, small improvements to the GDN prefill chunk kernels, bundled because they touch adjacent files in the same subsystem and independently deliver small measured wins. Each is its own commit so a reviewer can accept / revert individually.

| Commit | File | What |
|---|---|---|
| `solve_tril: relocate memset` | `layers/fla/ops/solve_tril.py` | Merge the destination-buffer pre-zero into the inversion kernel's first store. Bit-identical, one pass instead of two |
| `chunk-indices fused kernel` | `layers/fla/ops/index.py` + `v1/attention/backends/gdn_attn.py` | Replace the host-side Python loop + `.tolist()` (which forces a D2H sync per forward) with one fused Triton launch |
| `chunk_o: V-tile dedup + fp32-nocast dot` | `layers/fla/ops/chunk_o.py` | Compute the intra-chunk score matrix once, reuse across V-tiles. Drop a vestigial `fp32 → fp16` narrowing before the final FFMA dot (accumulator is already fp32) |

4 files, +198 / −34 lines total. No new dependencies. Kill switches / env flags: none — every change is bit-identical (chunk-indices, solve_tril) or precision-improving (chunk_o) vs the current code.

## Test Plan

Hardware: SM 7.0 (Tesla V100 SXM2 32GB), TP4.

1. **Bit-identical checks (solve_tril, chunk-indices):**
   - md5-compare kernel output vs the pre-patch path on fixed input seeds
   - Confirm the memset invariant still holds (the consumer's unmasked `tl.dot` reads the full block)
2. **Numerical assertivity (chunk_o):**
   - `rel_rms` against a CPU float32 reference (with a float64 anchor at ~2e-7)
   - Target: no arm systematically worse than the current path; expect chunk_o path *tighter* (fewer casts)
3. **Delivered A/B on served GDN prefill** — TTFT and prompt tok/s across rungs, boot-alternated repeats to guard against drift.

## Test Result

**`solve_tril` (bit-identical):**
- Output byte-for-byte equal to pre-patch on all tested shapes
- Measured faster at every shape (one pass instead of two; small but nonzero on the hot path)

**`chunk-indices` (bit-identical):**
- Index table byte-for-byte equal to the Python-loop path
- Per-forward D2H sync eliminated
- Measured faster in the served GDN prefill (30/40 layers of the target production model)

**`chunk_o`:**
- `rel_rms` vs fp32 reference: tighter than pre-patch on every tested shape; no arm systematically worse
- Delivered TTFT A/B on served large prefill: two-sided directional signal in the patch's favor
- No regression on smaller prefill rungs or on decode

All three shipped in a downstream fork's release earlier this month. Bundling here to send them upstream.

---

<details>
<summary>Essential Elements of an Effective PR Description Checklist</summary>

- [x] The purpose of the PR
- [x] The test plan
- [x] The test results
- [ ] (Optional) Documentation update — happy to add notes if reviewers want any
</details>

### DIFF
--- a/vllm/model_executor/layers/fla/ops/chunk_o.py
+++ b/vllm/model_executor/layers/fla/ops/chunk_o.py
@@ -103,7 +103,11 @@ def chunk_fwd_kernel_o(
     USE_G: tl.constexpr,
     IS_VARLEN: tl.constexpr,
 ):
-    i_v, i_t, i_bh = tl.program_id(0), tl.program_id(1), tl.program_id(2)
+    # patches/0001: de-duplicated V-tile score matrix. Grid dim 0 (was `cdiv(V,BV)`) is collapsed;
+    # b_A = q·kᵀ is built ONCE per (chunk, K-head) and the V-tiles are looped INTERNALLY, instead of
+    # rebuilding the identical b_A in every V-tile program. Bitwise-identical to the prior kernel
+    # (same fp32-accumulate order, same casts, same gating expressions), +12–22% on GDN prefill.
+    i_t, i_bh = tl.program_id(0), tl.program_id(1)
     i_b, i_h = i_bh // H, i_bh % H
 
     if IS_VARLEN:
@@ -130,9 +134,8 @@ def chunk_fwd_kernel_o(
     o += (bos * H + i_h) * V
     h += (i_tg * H + i_h).to(tl.int64) * V * K
 
-    b_o = tl.zeros([BT, BV], dtype=tl.float32)
+    # ---- intra-chunk score matrix b_A = q·kᵀ, built ONCE (V-tile-independent) ----
     b_A = tl.zeros([BT, BT], dtype=tl.float32)
-
     for i_k in range(tl.cdiv(K, BK)):
         p_q = tl.make_block_ptr(
             q, (T, K), (Hg * K, 1), (i_t * BT, i_k * BK), (BT, BK), (1, 0)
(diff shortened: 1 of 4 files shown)


yangzhuxinyzx:
Maintainer V100 audit found two blockers, so this PR is being returned to Draft rather than merged.

1. `prepare_chunk_indices_fused` is not bit-identical for nondecreasing `cu_seqlens` containing an empty segment. Reproducer on V100:

```python
cu = torch.tensor([0, 1, 65, 65, 100000], dtype=torch.int32)
off = prepare_chunk_offsets(cu, 64)
got = prepare_chunk_indices_fused(off, off.cuda()).cpu()
want = prepare_chunk_indices(cu, 64)
assert torch.equal(got, want)  # fails from chunk row 2 onward
# got sequence id 3, legacy/reference sequence id 2
```

The legacy implementation compresses empty segments; the candidate writes raw program ids. Either preserve that contract or prove/assert that this metadata path can never contain an empty segment, with CUDA-graph/mixed-batch regression coverage. Single positive-length sequences pass through 262144 tokens / 4096 chunks, so the dynamic chunk loop itself is good.

2. Ruff reports five E501 failures in `gdn_attn.py:1635-1639`.

The other evidence is promising: at B1/S1024, every saved intermediate through `stage_final_state` is bit-identical; only the intended chunk_o output changes (max abs 3.81e-6), and the full chunk pipeline measured 1.110 -> 0.984 ms (-11.3%). Please fix/split the index commit and add boundary tests; solve_tril and chunk_o can then be reconsidered independently.

kkobold:
Withdrawing — this PR went out before it cleared our internal review/validation bar. Apologies for the noise; will resubmit if/when it's actually ready.

PR #202, opened 2026-08-13
[Kernel][SM70] FA_V100 bundle: decode grid overlaunch + paged prefill sS/sP race fix

kkobold:
Two independent SM70 FA_V100 kernel fixes. Bundled as separate commits so either can be dropped if a reviewer wants only one.

Both patches were qualified on V100 SM70 (CUDA 12.8) with real-kernel A/B measurements. Numbers below are per-kernel operator-level (not end-to-end served throughput) unless otherwise noted.

---

### 1. `e4c8d1d` — decode: size launch grid by runtime seq_len, not workspace capacity

Fixes an eager over-launch in the paged FA_V100 decode-attention path.

`flash_attn_v100.py` was passing `workspace_seq_capacity_hint = block_table.shape[1] * block_size` — i.e. `max_model_len` worth of blocks — and `flash_attn_interface.py` then launched `workspace_num_partitions` rather than `runtime_num_partitions`. At `--max-model-len 262144` with `partition_size 256`, that is 1024 launched partitions where roughly 17 do any work.

The fix separates the two roles:
- `workspace_num_partitions` — upper bound, stays as scratch size and as the `max_num_partitions` template arg the kernels clamp with.
- `runtime_num_partitions` — grid.z at launch, sized from the effective `runtime_max_seq_len`.

Correctness preserved: both `flash_decode_paged` and the reduce kernel still receive the workspace bound and clamp coverage internally.

Measured speedups (V100 SM70, interleaved medians):

| rows | path | scalar over→fix | XQA (D=256,q_per_kv=6) |
|---|---|---|---|
| 32  | graph      | 866 → 518 µs (1.67x)    | — |
| 128 | graph edge | 3255 → 1756 µs (1.85x)  | 1667 → 1270 µs (1.31x) |
| 256 | eager      | 6408 → 3455 µs (1.85x)  | 3203 → 2394 µs (1.34x) |
| 512 | eager      | 13034 → 6839 µs (1.91x) | 6129 → 4507 µs (1.36x) |

At long context (S ≥ 32768) the partition size rises to 1024, shrinking over-launch to ~256 partitions and the gain to ~2%. Small-B graph-captured shapes are already bucketed → ≈0 change. No shape regresses. Same speedup ratio device-confirmed on the `fp8_e5m2` KV route.

Files: `flash-attention-v100/flash_attn_v100/flash_attn_interface.py`, `vllm/v1/attention/backends/flash_attn_v100.py`.

---

### 2. `8aa6b24` — paged prefill: two-phase P commit fixes sS/sP union race (D≠256)

Fixes a WAR/RAW race in `flash_attention_forward_kernel_paged` on head_dim ≠ 256.

`fused_mha_forward_paged.cu` declares `union reuse_sp { float s[]; __half p[]; }`. For D≠256, both `sS` and `sP` point at the same bytes. In the softmax sub-tile the loop reads sS as `float4` and, inside the same tid-guarded region, commits P to sP as `__half2`/`__half`, with only within-thread ordering via a register buffer. Cross-warp, warp W's P store lands on the float score row that warp W' is still reading as float4 — no ordering between them.

Observed on baseline (compute-sanitizer racecheck): **4 errors at D=128, 5 at D=64**; 10 identical-input launches at D=128 H=8 give **10 unique outputs**. D=256 is immune (sP is a separate `p_strict` buffer).

Fix: split the guarded region into (A) all sS reads + reductions with P staged in registers, and (B) all sP stores, with one uniform `__syncthreads()` between them at block-uniform control flow. The whole transform lives under `if constexpr (D != 256)`, so the D=256 codegen is unchanged.

Measured cost (real-kernel A/B, interleaved):
- **D=256 (immune, served path): ZERO cost** — all 39 D=256 kernel symbols are raw-SASS byte-identical to baseline; `cuobjdump -res-usage` identical.
- **D=64 short-context** (M=512/KV=4096): +0.5% (426.0 → 428.1 µs) — a fixed barrier cost, vanishes proportionally at higher load (D=64 M=1024/KV=8192: +0.02%, noise).
- **D=128** (the head the race actually fires on): -0.07% / -0.3% at 512/4096 and 1024/8192 — inside noise, so honestly "D=128 pays nothing".

Correctness: racecheck 4→0 (D=128), 5→0 (D=64). Determinism D=128 H=8: 10/10-unique baseline → 1 unique (bitwise-identical across 10 runs). Bit-equal to baseline on all previously-stable shapes — the fix reorders **when** threads read, never **what** is computed.

Note on prior art: this is the surgical form of an earlier internal patch that hoisted variables for every D. That earlier form measured +0.47…+2.79% D=256 cost from register/spill perturbation on the immune path. Guarding the transform under `if constexpr (D != 256)` makes the D=256 machine code literally unchanged, at no cost to the fix on the D≠256 paths.

Files: `flash-attention-v100/kernel/fused_mha_forward_paged.cu`.

---

### Testing on our side

Both patches shipped in monico-vllm v0.1.4/v0.1.6 on 4× V100 SXM2, serving at TP4 with `FLASH_ATTN_V100` backend. In-engine A/B with the resident model (head_dim 256, AWQ int4, fp8_e5m2 KV) is stable; racecheck-clean device runs on D=128/D=64 shapes for #2.

Happy to split into two PRs if that's easier to review.

### DIFF
--- a/flash-attention-v100/flash_attn_v100/flash_attn_interface.py
+++ b/flash-attention-v100/flash_attn_v100/flash_attn_interface.py
@@ -213,6 +213,17 @@ def _get_decode_plan(
         ),
         workspace_num_partitions=workspace_num_partitions,
     )
+    # Invariant, by construction: actual <= launch <= workspace. Both partition
+    # counts divide by the same positive partition_size, and
+    # effective_workspace_seq_capacity is floored to >= effective_max_seq_len
+    # above (:164-168), so workspace_num_partitions >= runtime_num_partitions;
+    # launch_num_partitions is one of those two. A check of that inequality HERE
+    # would be a tautology that can never fire (it compares three quantities all
+    # derived from the same hints), so it is NOT placed here. The coverage that
+    # actually matters -- launch >= ceil(true seq_lens.max()/ps) -- can only be
+    # checked where the real per-sequence lengths are known, which is the caller
+    # wrappers flash_attn_decode_paged / _xqa; see
+    # _assert_decode_launch_covers_seq_lens.
     device_index = q.device.index if q.device.index is not None else -1
     key = (
         device_index,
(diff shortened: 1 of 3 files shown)


yangzhuxinyzx (approved):
Local V100/SM70 audit passed on CUDA 12.8 / Torch 2.10:\n\n- built both extensions with TORCH_CUDA_ARCH_LIST=7.0\n- 52 SM70 backend policy tests passed\n- decode planner: 8 passed; MTP small-q: 2 passed; varlen layout: 18 passed\n- reproduced baseline D128 nondeterminism (10 unique hashes / 10 launches) and candidate stability (1 / 10) at M512/S4096 and M1024/S8192\n- interleaved kernel A/B: D128 effectively flat, D64 about +0.7%, D256 flat\n- runtime-grid cap: 1.076 -> 0.707 ms at 32 rows and 3.430 -> 2.412 ms at 128 rows\n- adversarial under-coverage guard rejects before launch\n\nAccepted for main.

PR #203, opened 2026-08-13
[Kernel][SM70] AWQ/FP8 MoE per-expert dispatch: size CTA tile from per-expert M, not total_slots

kkobold:
Fixes the per-expert AWQ/FP8 MoE GEMM tile heuristic on SM70.

### Problem

`awq_moe_gemm_sm70_per_expert_dispatch_out` sets `op.dispatch_num_override = 1`. `Gemm::Run` then builds a dispatch-only context with `num=1` while the real launch keeps the grouped descriptors (`num=num_experts`, `expert_offsets`). Because `Context::Filter` keys the tile purely on `batch_size = desc.m` independent of `num`, the heuristic sees a single GEMM of `m = total_slots, num=1` and picks a big-`m` tile — while each expert actually computes only `total_slots/num_experts` rows. The M dimension of the tile is grossly over-sized.

At B=128 decode (active ≈ 251/256, ~4 rows/expert on Ornith-1.0-35B) the default tile's cost balloons — `w13` default = 0.96 ms at active=256 vs 0.147 ms at active ≤ 32, a 6.5x cliff.

### Fix — capture-safe

Add `dispatch_m_override` (types.h) and extend the existing override block in `gemm.cu` to also rewrite `desc.m` on the dispatch context. At the two op sites, a representative per-expert M is computed **host-side ahead of graph capture** from the routing counts and passed in.

### Why "capture-safe"

An earlier attempt computed the representative M in C++ at the dispatch site via `counts.max().item<int>()`. That `.item()` is a synchronizing D→H copy and it runs inside the MoE decode forward, which vLLM V1 captures into full CUDA graphs — it threw `cudaErrorStreamCaptureInvalidated` and the engine could not boot. This variant computes the override in Python before entering the graph.

### Delivered numbers

Per-op median-of-round-medians (real grouped launch, 256 experts, Ornith shapes, decode band):

| slots | ≈B | w13 def→determ | w2 def→determ |
|---|---|---|---|
| 512  | 64  | **1.83x** | 1.37x |
| 1024 | 128 | **1.66x** | 2.66x |

End-to-end delivered (v0.1.0 → v0.1.1, served config, quality unchanged):
- decode tok/s **+2.5..+5.4%** at B=4..64, ties at B=1
- TTFT **+30..45%** at B=4..32 — countercurrent; the fix picks a splitK route that costs prefill setup time
- output quality: fp64 ledger inherited from a sibling patch shows every pick has identical deviation from fp32 truth (quality-safe)

Ships as monico-vllm v0.1.1.

### Files

- `csrc/ops.h`, `csrc/torch_bindings.cpp` — signature update
- `csrc/sm70_turbomind/lmdeploy/src/turbomind/kernels/gemm/types.h`, `.../gemm.cu` — the `dispatch_m_override` plumbing
- `csrc/sm70_turbomind/ops/awq_sm70_gemm.cu` — pass through to the dispatch context
- `vllm/_sm70_ops.py`, `.../quantization/awq_sm70_moe.py`, `.../quantization/fp8_sm70_moe.py` — host-side computation of the representative M

Origin: monico-vllm `patches/0029-awq-moe-per-expert-tile-sizing`.

### DIFF
--- a/csrc/ops.h
+++ b/csrc/ops.h
@@ -315,7 +315,8 @@ void awq_moe_gemm_sm70_per_expert_dispatch_out(torch::Tensor out,
                                                int64_t k,
                                                int64_t n,
                                                int64_t group_size,
-                                               bool gated_silu);
+                                               bool gated_silu,
+                                               int64_t dispatch_m_override = 0);
 
 void awq_moe_dense_stage_sm70_out(torch::Tensor out,
                                   torch::Tensor input,
@@ -476,7 +477,8 @@ void fp8_moe_gemm_sm70_per_expert_dispatch_out(torch::Tensor out,
                                                int64_t k,
                                                int64_t n,
                                                int64_t group_size,
-                                               bool gated_silu);
+                                               bool gated_silu,
+                                               int64_t dispatch_m_override = 0);
 
 void fp8_moe_dense_stage_sm70_out(torch::Tensor out,
                                   torch::Tensor input,
--- a/csrc/sm70_turbomind/lmdeploy/src/turbomind/kernels/gemm/gemm.cu
+++ b/csrc/sm70_turbomind/lmdeploy/src/turbomind/kernels/gemm/gemm.cu
@@ -655,13 +655,34 @@ int Gemm::Run(const Operation&    operation,
 
     std::optional<Context> dispatch_context_storage;
     Context*               dispatch_context = &context;
-    if (operation.dispatch_num_override > 0 && operation.dispatch_num_override != context.desc().num) {
+    const bool want_num_override =
+        operation.dispatch_num_override > 0 && operation.dispatch_num_override != context.desc().num;
+    // issues/0027 / patches/0029: for a grouped MoE launch the num_override path
+    // rewrites only desc.num, leaving desc.m = total_slots, so Context::Filter
+    // (context.cu:134, keys on desc.m) sizes the CTA tile for the grouped total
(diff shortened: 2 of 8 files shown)


yangzhuxinyzx:
Maintainer audit found blocking FP8 wiring errors; returning this PR to Draft.

1. `fp8_moe_gemm_sm70_out()` passes `dispatch_m_override` even though the function has no such parameter or local. Ruff reports F821 at `vllm/_sm70_ops.py:1656`; every ordinary FP8 batched call reaches a Python `NameError` before the op.

2. `fp8_moe_gemm_sm70_per_expert_dispatch_out()` accepts `dispatch_m_override` but does not pass it to `torch.ops._C`. The C++ schema default remains 0, so the advertised FP8 optimization is disabled.

3. The fake registrations are asymmetric with the schemas: the ordinary FP8/AWQ fakes gained an override argument even though those ordinary schemas did not, while the FP8 per-expert fake did not gain the new optional argument.

Please repair the wrapper/fake signatures and add tests that call both ordinary and per-expert AWQ/FP8 wrappers, then run V100 exactness plus CUDA-graph capture for balanced, random, and single-hot-expert routing. The representative average-M heuristic also needs skewed-routing timing evidence because `total_slots // num_experts` can be far below the hottest expert's M.

kkobold:
Withdrawing — this PR went out before it cleared our internal review/validation bar. Apologies for the noise; will resubmit if/when it's actually ready.

PR #239, opened 2026-08-18
[Frontend][Model] Per-model serving defaults, DeepSeek-OCR resolution modes, and SWA async-eviction fix (SM70)

kkobold:
## What

Serving-quality improvements for long-generation OCR workloads, plus a
correctness fix for the prefix-anchored SWA capability merged in #281:

1. **Bugfix (SWA/#281):** out-of-window gap blocks were freed on the
   scheduled-token basis; under async scheduling this can free blocks the
   worker has not yet computed past. Free on the processed-token basis
   instead. Regression tests simulate the async gap.
2. **Per-model serving defaults with request-override bounds:** an
   architecture-keyed registry applied at the chat endpoint. A model can
   declare its recommended sampling/extra-args recipe; requests may
   override within declared bounds. DeepSeek-OCR registers its checkpoint
   README recipe (n-gram blocker 30/90 with whitelist tokens, temperature
   0, max_tokens 8192) so an unconfigured client gets the model's intended
   behavior via plain chat completions.
3. **Architecture-keyed chat template fallbacks:** DeepSeek-OCR's
   model_type previously fell through to the DeepSeek-VL2 template, which
   injects role markers the OCR checkpoint was not trained with; the
   fallback now pins the raw-concatenation prompt (golden test included).
4. **Repetition-detection default** for OCR serving: a bounded
   exact-pattern detector as a safety net for degenerate table pages,
   request-overridable.
5. **Per-request DeepSeek-OCR resolution modes:** the checkpoint's five
   official modes (tiny/small/base/large/gundam) selectable via
   `mm_processor_kwargs: {"image_mode": ...}`; multi-image serving
   profile; `max_crops` plumbing. The serving default remains the
   checkpoint's gundam configuration.

## Why

On document OCR, the model's own recipe is load-bearing: without the
n-gram blocker and the correct raw prompt, long pages degenerate into
loops. These changes make the endpoint serve the checkpoint's intended
recipe by default while keeping every knob per-request-overridable within
bounds. The SWA fix closes an async-scheduling correctness gap in the
capability added by #281.

## Validation

All device work on Tesla V100-SXM2 32 GB (sm_70), fp16, on current main.

- 195 CPU tests green, including #281's admission/mask/manager suites
  (proves coexistence with the merged SWA capability), the new
  eviction-basis regression tests, resolution-mode suite, template golden,
  defaults/bounds and repetition-detection suites.
- Endpoint (serve) on V100: deterministic 3× byte-identical generations;
  resolution-mode discrimination (default ≡ gundam as identity control;
  base/large produce distinct image-token counts); document-page scoring
  against known ground truth (per-mode table in tests receipt).

## Serving DeepSeek-OCR (validated configuration)

Launch line used for every endpoint validation in this PR (Tesla
V100-SXM2 32 GB, fp16):

```bash
vllm serve /path/to/DeepSeek-OCR \
  --dtype float16 \
  --max-model-len 8192 \
  --gpu-memory-utilization 0.85 \
  --enforce-eager \
  --no-enable-prefix-caching \
  --mm-processor-cache-gb 0 \
  --logits-processors vllm.model_executor.models.deepseek_ocr:NGramPerReqLogitsProcessor
```

- `--logits-processors` loads the n-gram blocker class; the serving-defaults
  registry supplies its per-request parameters and **fails startup with the
  exact flag text if the class is missing** (never silent looping).
- `--no-enable-prefix-caching` / `--mm-processor-cache-gb 0` are the
  checkpoint recipe's engine flags.
- For multi-image requests add `--limit-mm-per-prompt '{"image":8}'`
  (the SM70 baseline defaults to 1 image per prompt).
- `--enforce-eager` is the validated configuration; CUDA graphs measured
  ~4x faster decode at ~1% window cost and can be enabled after a
  graph-mode validation pass.
- Tensor parallelism: the checkpoint has 10 attention heads, so
  `--tensor-parallel-size` must divide 10 (1 or 2 on a 4-GPU host); 4
  fails at startup with the head-divisibility error.

## Merge base

Branch applies and merges cleanly onto current `main` at `34403018d`
([Kernel][SM70] Optimize E4M3 page800 long decode, #290); all validation in
this PR ran on that base.

### DIFF
--- a/flash-attention-v100/flash_attn_v100/flash_attn_interface.py
+++ b/flash-attention-v100/flash_attn_v100/flash_attn_interface.py
@@ -897,6 +897,8 @@ def flash_attn_decode_paged(
     workspace_seq_capacity_hint: int | None = None,
     active_num_partitions: torch.Tensor | None = None,
     partition_size_hint: int | None = None,
+    anchor_lens: torch.Tensor | None = None,
+    anchored_window: int = 0,
 ):
     if softmax_scale is None:
         softmax_scale = q.shape[-1] ** -0.5
@@ -905,9 +907,12 @@ def flash_attn_decode_paged(
     block_table = maybe_contiguous(block_table)
     seq_lens = maybe_contiguous(seq_lens)
     out = maybe_contiguous(out)
+    anchor_lens = maybe_contiguous(anchor_lens)
     window_size_left, window_size_right = window_size
     if window_size_left < -1 or window_size_right < -1:
         raise ValueError(f"Invalid window_size={window_size}; values must be >= -1")
+    if anchored_window > 0 and anchor_lens is None:
+        raise ValueError("anchored_window requires anchor_lens")
     batch_capacity = q.shape[0]
     num_heads = q.shape[1]
     head_dim = q.shape[2]
@@ -960,6 +965,8 @@ def flash_attn_decode_paged(
         float(v_scale),
         int(window_size_left),
         int(window_size_right),
+        anchor_lens,
+        int(anchored_window),
     )
 
 
@@ -1232,6 +1239,8 @@ def flash_attn_prefill_paged(
     v_scale: float = 1.0,
     causal: bool = True,
     window_size: tuple = (-1, -1),
+    anchor_lens: torch.Tensor | None = None,
+    anchored_window: int = 0,
 ):
     if softmax_scale is None:
         softmax_scale = q.shape[-1] ** -0.5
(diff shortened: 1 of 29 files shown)


yangzhuxinyzx (changes requested):
Thanks for the careful kernel and manager work. I found two correctness blockers in the integration path:

1. The anchored metadata is only attached in FlashAttnV100MetadataBuilder.build_for_cudagraph_capture(). The normal build() path calls _attach_common_flash_metadata(), but that helper does not copy common_attn_metadata.prefix_anchor_lens or set decode_sliding_window; build_for_drafting() has the same omission. TritonAttentionMetadataBuilder.build() does not propagate arbitrary CommonAttentionMetadata fields. A direct source-level probe with CommonAttentionMetadata(prefix_anchor_lens=[16]) produced regular Flash metadata with neither prefix_anchor_lens nor decode_sliding_window. Therefore _anchored_swa_params() returns (None, 0) on ordinary eager/PIECEWISE runtime metadata. This PR itself downgrades FULL graphs to PIECEWISE, so the KV manager can evict the gap while the attention path runs unmasked. Please attach the persistent anchor buffer/window in every runtime metadata path (preferably through one shared helper), including drafting if supported, and add an integration test that exercises builder.build(), not only direct kernel calls.

2. decode_sliding_window is activated on non-None but never validated as positive. PrefixAnchoredSWAManager uses 0/negative values in its eviction arithmetic, while _anchored_swa_params() explicitly disables the mask for values <= 0. Repro with block_size=4, prompt=16, computed=28: window=0 evicts logical blocks [4,5,6] and window=-8 evicts [4,5,6,7,8], yet the backend reports mask inactive for both. Please reject non-integer/non-positive values before constraints/spec construction and add zero/negative validation tests.

Local audit results otherwise: merge against current main was clean; git diff --check passed; Ruff 0.14.0 check/format passed on all 26 changed Python files; the new mask and single-type KV-manager suites passed 14/14. I did not repeat the expensive model E2E evidence.

PR #247, opened 2026-08-19
[Bugfix][KV Offloading] Backport mamba/GDN CPU offloading fixes (upstream vllm#42554 + vllm#44599)

lwh9346:
# [Bugfix][KV Offloading] Backport mamba/GDN CPU offloading fixes (upstream vllm#42554 + vllm#44599)

## Purpose

Hybrid GDN models (Qwen3.8 / Qwen3-Next family, `linear_attention` + periodic
`full_attention` layers) are broken with the native CPU KV offloading connector
(`--kv-offloading-size N --kv-offloading-backend native`) as soon as any prefix
is repeated:

1. **Engine crash on offload hit.** `Scheduler._mamba_block_aligned_split`
   asserted `num_external_computed_tokens == 0` ("External KV connector is not
   verified yet"). With offloading enabled, any request whose prefix hits the
   CPU offload pool reaches the scheduler with external computed tokens and
   immediately kills the engine.
2. **Misaligned hit windows.** `OffloadingConnectorScheduler` treated
   `MambaSpec` groups like sliding-window attention when computing hit windows.
   In `align` cache mode a mamba group keeps a single state per block, so a hit
   window that is not a multiple of the offloaded block size is incorrect.

This PR backports the two upstream fixes:

- [vllm-project/vllm#42554](https://github.com/vllm-project/vllm/pull/42554)
  (merged upstream as `68f5e565`): drop the assertion; skip the mamba-aligned
  split when `load_kv_async` (async receive setup does no local work).
- [vllm-project/vllm#44599](https://github.com/vllm-project/vllm/pull/44599)
  (merged upstream as `b927004c`): add `resolve_mamba_align_size()` and round
  the hit window down to the mamba-aligned offloaded block size.

**Drop this commit when rebasing onto an upstream base that contains both PRs.**

## Test Plan / Result

`tests/test_gdn_offload.py` (1Cat workspace) on Qwen3.8-27B-FP8, 8xV100, TP4,
`--kv-offloading-size 64 --kv-offloading-backend native`:

| step | result |
|---|---|
| cold request (long shared prefix) | 10.7s TTFT |
| reset prefix cache, same prefix via CPU offload hit | **2.2s TTFT** |
| greedy outputs cold vs offload-hit | **byte-identical** |
| follow-up distinct requests | engine healthy |

Without the patch, step 2 crashes the engine at the scheduler assertion.

PR #248, opened 2026-08-19
[Bugfix][KV Offloading] Allow native CPU offload with expandable_segments

Leonccaa:
## Purpose

Allow 1Cat's native `OffloadingConnector` to run with
`PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` without weakening the
default safety check for other KV connectors.

The native CPU offload path resolves the current tensor virtual address for
every transfer and uses `cuMemcpyBatchAsync` / `cudaMemcpyAsync`. It does not
register or export vLLM KV device memory through RDMA, GDS, or CUDA IPC, so a
CUDA VMM physical-page remap does not invalidate its transfer state.

This change:

- adds an explicit `SupportsVmmSafeTransfers` connector capability;
- lets the config validator accept only connectors that opt into that
  capability; and
- marks the native `OffloadingConnector` as VMM-safe.

All unmarked connectors remain fail-closed exactly as before.

## Scope and duplicate-work check

This PR is intentionally limited to allocator/connector compatibility. It does
not include the separate MTP/Mamba external-hit correctness fix, submitted as
1Cat #249.

Upstream vLLM PR
[#42405](https://github.com/vllm-project/vllm/pull/42405) proposes the same
connector-declared marker for `SimpleCPUOffloadConnector`; PR
[#48243](https://github.com/vllm-project/vllm/pull/48243) proposes an
operator-level opt-out. Neither is merged, and no open 1Cat PR covers 1Cat's
native `OffloadingConnector`. This PR applies the fail-closed connector
capability design to that native path and includes 1Cat/V100 validation.

## Test Plan

```bash
.venv/bin/python -m pytest -q tests/v1/kv_connector/unit/test_config.py
pre-commit run --from-ref upstream/main --to-ref HEAD
git diff --check upstream/main..HEAD
```

## Test Result

- Config tests: **14 passed**.
- All pre-commit hooks for the five changed files passed.
- `git diff --check` passed.
- Integrated hardware validation used the identical five-file diff, layered
  with 1Cat PR #247, on 4 x Tesla V100 32 GB with native CPU KV offload and
  `expandable_segments:True`. PIECEWISE and FULL CUDA Graph capture completed;
  a cold A / cold B / external-hit A sequence completed with byte-identical A
  output and real GPU-to-CPU plus CPU-to-GPU transfers. The prior
  `static_cuda_launcher` / `CUDA driver error: invalid argument` failure did
  not recur.

## AI assistance

AI assistance was used for implementation, test orchestration, and drafting
this PR. The human submitter has reviewed every changed line and is submitting this
PR for maintainer review.

### DIFF
--- a/tests/v1/kv_connector/unit/test_config.py
+++ b/tests/v1/kv_connector/unit/test_config.py
@@ -104,6 +104,16 @@ def test_kv_connector_rejects_expandable_segments(monkeypatch, kv_connector):
         _build_config(kv_connector=kv_connector)
 
 
+def test_native_offloading_allows_expandable_segments(monkeypatch):
+    """Native CPU offload resolves GPU virtual addresses for each CUDA copy
+    and explicitly opts into the VMM-safe transfer contract."""
+    monkeypatch.setenv("PYTORCH_CUDA_ALLOC_CONF", "expandable_segments:True")
+
+    cfg = _build_config(kv_connector="OffloadingConnector")
+
+    assert cfg._connector_supports_vmm_safe_transfers()
+
+
 def test_kv_connector_allows_expandable_segments_with_sleep_mode(monkeypatch):
     """Sleep mode routes KV allocations through CuMemAllocator's pool, which
     auto-disables expandable_segments (see #40812)."""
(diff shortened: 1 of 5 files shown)


yangzhuxinyzx (approved):
源码审计通过：VMM-safe capability 为显式 opt-in，未知/RDMA connector 保持 fail-closed；native OffloadingConnector 的 GPU 传输每次解析 tensor 虚拟地址并仅使用 CUDA memcpy。当前 main 合并态及 diff/语法/Ruff/格式检查通过。

PR #249, opened 2026-08-19
[Bugfix][KV Offloading] Preserve MTP/Mamba state across external hits

Leonccaa:
## Purpose

Fix incorrect native CPU KV offload restores for hybrid Mamba/GDN models when
MTP speculative decoding is enabled.

This PR **depends on 1Cat PR #247**. PR #247 removes the original external-KV
assertion and backports the Mamba boundary-alignment fixes from upstream vLLM,
but it does not cover MTP-owned accepted-token state or mutable Mamba `align`
source blocks. This branch is stacked on the fixed head of #247
(`c514f39e15ca74a1f51bb5270f0f4734d1da8bc7`) and will be rebased after #247
lands.

## Root cause

On a real external CPU-tier hit, MTP4 output consistently lost exactly four
tokens. The fixed loss matching the MTP width exposed two related ownership
bugs:

1. Asynchronous accepted-token D2H copies targeted mutable `InputBatch` rows.
   Rows can be removed, condensed, resumed, or reused before the event is
   consumed, so accepted-token state could be applied to the wrong request.
2. A Mamba `align` table is not append-only. MTP scratch blocks move to new
   logical positions when a boundary is committed, while the native connector's
   block mirror only appends. A positional store could therefore save a stale
   Mamba state under a valid prefix key.

## Changes

- Store accepted-token D2H results in runner-owned buffers and remap them by
  request identity after synchronization. New, resumed, or same-ID replacement
  requests reset their speculative state.
- Clip attention block tables at the computed-token frontier without clipping
  allocator-defined non-attention state groups.
- Emit exact core-selected Mamba `align` boundary block IDs to the connector.
- Exclude mutable Mamba groups from the normal positional store path and build
  their store jobs only from exact boundary handoffs.
- Keep the existing pending-job fence on each source block until the async
  store completes, including deduplication when MTP aliases one physical block.

## Scope and duplicate-work check

This is not duplicate work:

- 1Cat #247 is the required base fix; it does not address MTP state ownership
  or the exact external-hit corruption described here.
- Upstream vLLM
  [#51358](https://github.com/vllm-project/vllm/pull/51358) implements exact
  Mamba boundary persistence for Mooncake with its own pin/watermark protocol;
  this PR adapts the correctness invariant to the native `OffloadingConnector`
  and its per-job fence lifecycle.
- Upstream vLLM
  [#52771](https://github.com/vllm-project/vllm/pull/52771) fixes MTP/EAGLE
  lookups being reduced to zero. In this reproduction the external hit was
  already real; the restored recurrent state was wrong.
- Upstream vLLM
  [#52832](https://github.com/vllm-project/vllm/pull/52832) concerns
  Mooncake finish-time partial tails and does not cover this native exact-
  boundary path.

No open 1Cat PR or issue was found for this mechanism.

## Test Plan

```bash
.venv/bin/python -m pytest -q \
  tests/v1/core/test_prefix_caching.py \
  tests/v1/kv_connector/unit/offloading_connector/test_scheduler.py \
  tests/v1/core/test_scheduler.py

.venv/bin/python -m pytest -q tests/v1/worker/test_gpu_model_runner.py \
  -k 'sync_mamba_accepted_token_state_tracks_request_ownership or mamba_align_postprocess_uses_runner_owned_d2h_buffers or mamba_align_cpu_postprocess_mirrors_runner_owned_snapshot or mamba_all_mode_uses_runner_owned_d2h_buffers'

pre-commit run --from-ref upstream/pr-247 --to-ref HEAD
git diff --check upstream/pr-247..HEAD
```

## Test Result

- Core, scheduler, and native connector regression suite: **219 passed**.
- New worker state-ownership tests: **4 passed**.
- All pre-commit hooks for the ten changed files passed.
- `git diff --check` passed.
- An extended run including the entire worker file produced **251 passed, 3
  failed**. All three failures reproduce unchanged on the unmodified #247 base:
  one existing sampling mock lacks `drafter`, and two GPU/FLASHINFER assertions
  are not valid in the CPU-only test environment.
- Integrated 4 x Tesla V100 validation of the identical MTP diff, together with
  #247 and the separately submitted VMM-safe connector change in 1Cat #248,
  passed a
  39,999-token cold / external-hit / local-rehit sequence with byte- and
  SHA-identical output. Exact-boundary, mid-block, Responses API, tool-call, and
  follow-up cases passed; MTP reported 136 drafted / 135 accepted tokens, with
  no OOM, process restart, or EngineCore fatal error.

## AI assistance

AI assistance was used for implementation, test orchestration, root-cause
analysis, and drafting this PR. The human submitter has reviewed every changed line and is submitting this
PR for maintainer review.

### DIFF
--- a/tests/v1/core/test_prefix_caching.py
+++ b/tests/v1/core/test_prefix_caching.py
@@ -1009,6 +1009,98 @@ def test_prefill_hybrid_model_mamba_align():
     manager.free(req0)
 
 
+def test_mamba_align_offload_handoff_tracks_relocated_mtp_boundary():
+    """Expose the committed state block, not its stale logical position.
+
+    A large second prefill chunk relocates the four MTP scratch blocks. The
+    first scratch block moves from logical position 1 to position 5 and becomes
+    the state after token 24. An append-only connector mirror would still see
+    that physical block at position 1 and could save the token-24 state under
+    the token-8 key.
+    """
+    block_size = 4
+    kv_cache_config = KVCacheConfig(
+        num_blocks=64,
+        kv_cache_tensors=[],
+        kv_cache_groups=[
+            KVCacheGroupSpec(
+                ["full"],
+                FullAttentionSpec(
+                    block_size=block_size,
+                    num_kv_heads=1,
+                    head_size=1,
+                    dtype=torch.float32,
+                ),
+            ),
+            KVCacheGroupSpec(
+                ["mamba"],
+                MambaSpec(
+                    block_size=block_size,
+                    shapes=((1, 1),),
+                    dtypes=(torch.float32,),
+                    mamba_cache_mode="align",
+                    num_speculative_blocks=4,
+                ),
+            ),
+        ],
+    )
+    manager = KVCacheManager(
+        kv_cache_config,
+        max_model_len=8192,
+        enable_caching=True,
+        hash_block_size=block_size,
+    )
+    request = make_request("0", [0] * 24, block_size, sha256)
+    computed_blocks, num_computed_tokens = manager.get_computed_blocks(request)
+
+    assert (
+        manager.allocate_slots(
+            request,
+            num_new_tokens=4,
+            num_new_computed_tokens=num_computed_tokens,
+            new_computed_blocks=computed_blocks,
+        )
+        is not None
+    )
+    mamba_manager = manager.coordinator.single_type_managers[1]
+    mamba_blocks = mamba_manager.req_to_blocks[request.request_id]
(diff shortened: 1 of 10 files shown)


yangzhuxinyzx (approved):
源码审计通过：accepted-token D2H 状态按请求身份安全重映射；attention lookahead 裁剪与非 attention 状态隔离；Mamba align 使用同一步 scheduler 的精确 boundary block handoff，并在异步 store 生命周期沿用 source-block fence。当前 main 合并态及 diff/语法/Ruff/格式检查通过。

PR #250, opened 2026-08-19
[Bugfix][SM70] Fix the DeepSeek-V4 indexer cache layout and restore the ReLU

dg1kjd:
## Purpose

Two independent defects in the SM70 lightning-indexer path. Both only bite once top-k is actually selective, i.e. at long context — which is why the 8k bring-up (#184, #159) looked clean.

### 1. The paged decode kernels read a block-major cache as token-interleaved

`indexer_k_quant_and_cache` / `cp_gather_indexer_k_quant_cache` (`csrc/libtorch_stable/cache_kernels.cu`) store, **per block**, all `block_size` rows of 128 FP8 values first and only then that block's FP32 scales:

```
value: block_base + pos_in_block * 128 + d
scale: block_base + block_size * 128 + pos_in_block * 4
```

Every Triton kernel in `sm70/indexer.py` instead assumed a per-token `[128 values][4-byte scale]` record at `pos_in_block * cache.stride(1)`:

```python
token_ptr = cache_ptr + physical_block.to(tl.int64) * cache_stride0 \
                      + pos_in_block * cache_stride1
scale_ptr = (token_ptr + head_dim).to(tl.pointer_type(tl.float32))
```

Only token 0 of each block read the right *values*, and **no** token read the right scale — what it picked up was four FP8 value bytes reinterpreted as FP32. Measured in situ: 6943 of 13802 scales negative, spanning ±1e36..1e38.

Prefill was unaffected because it gathers through the layout-aware C++ kernel, so retrieval looked perfect there (needle at rank 0–16 of 13801) while decode was broken (rank 43–3323, **0 of 8** needle keys kept). That asymmetry is what made this look like "long-context quality decay" rather than a bug.

Fixed with a shared `_index_cache_ptrs()` helper so the layout is stated once.

### 2. The scoring function dropped the per-head ReLU

The model computes

```
I[t, s] = sum_h weights[t, h] * relu(q[t, h] . k[s])
```

(reference `Indexer.forward`, `inference/model.py:427`; the in-repo `fp8_mqa_logits` reference at `v1/attention/ops/rocm_aiter_mla_sparse.py:510` applies the same ReLU, and every non-SM70 backend applies it inside its logits kernel). SM70 computed

```
(sum_h weights[t, h] * q[t, h]) . k[s]
```

folding the weights out through a nonlinearity. That is 64× cheaper and a different function: on random data the induced top-512 sets agree only 51–66%.

### Result

Needle-in-a-haystack, ctx {4k, 16k, 64k} × depth {0.25, 0.5, 0.75}: **5/9 → 9/9 EXACT**. Extending the grid to ctx {128k, 240k} × the same depths gives a further **6/6 EXACT** after the fix (207,202-token prompts); no pre-fix baseline was taken at those two lengths, so that half is an after-only result. 15/15 EXACT from 4k to 207k with the fix in place.

### Making the ReLU affordable

The correct form is 64× the flops of the factored one, and Triton's `tl.dot` does not reach Volta's tensor cores: on this shape it measures **3.3 TFLOP/s** against **54** for cuBLAS fp16 and **13** for plain fp32 CUDA cores. No tiling fixes it — folding heads into M to get a `[256,128]x[128,64]` MMA measured 3.34. Prefill scoring therefore goes through `torch.mm` as one `[tokens*heads, 128] x [128, key_tile]` GEMM with a fused Triton `relu → weight → head-sum → key-scale` epilogue; keys are tiled to bound the score tile.

- Standalone: **3.39 → 18.1 TFLOP/s**
- In-engine 207k-token prefill: **768 s → 350 s** (128k: 268 → 148 s)
- Retrieval unchanged at 6/6 EXACT

Both scoring paths stay available for A/B: `VLLM_SM70_INDEXER_RELU=0` restores the factored form, `VLLM_SM70_INDEXER_PREFILL_CUBLAS=0` the Triton MMA.

## Test Plan

```bash
pytest tests/kernels/test_deepseek_v4_sm70_indexer.py -v
```

**`test_sm70_indexer_prefill_dequantizes_fp8_storage_as_uint8` had to be replaced, not extended.** Its expected value *was* the factored form:

```python
weighted_q = torch.sum(q.float() * weights[:, :, None], dim=1).half()
expected = torch.mm(weighted_q, k_reference.T, out_dtype=torch.float32)
```

i.e. the test encoded defect 2 as its reference, so restoring the ReLU necessarily turns it red. The replacement scores against `sum_h w * relu(q . k)` directly. It also keeps FP8 keys as exact bit patterns (a small palette: `0x38 = 1.0`, `0xB8 = -1.0`, …) so the reference needs no native FP8 conversion, which SM70 lacks — the property the old test was written for is preserved.

Six tests, all on the SM70 gate already used by this file:

| test | pins |
|---|---|
| `test_prefill_matches_the_reference_scoring_function[True/False]` | both prefill paths (cuBLAS and Triton MMA) against `sum_h w·relu(q·k)`; also asserts the factored form is *not* within tolerance, so the ReLU cannot be silently dropped again |
| `test_prefill_relu_off_restores_the_factored_form` | the `VLLM_SM70_INDEXER_RELU=0` A/B escape hatch still works |
| `test_decode_reads_the_block_major_paged_cache[relu / factored-fused / factored-gather]` | all three decode paths, against a cache built in the **real** block-major layout with **distinct per-token scales** and a shuffled block table |

The decode test is what pins defect 1: with one scale per block, or an identity block table, the layout bug is invisible.

## Test Result

Hardware: 8× Tesla V100-SXM2-32GB (DGX-1), CUDA 12.9, Python 3.12. One GPU.

```
test_prefill_matches_the_reference_scoring_function[True]      PASSED
test_prefill_matches_the_reference_scoring_function[False]     PASSED
test_prefill_relu_off_restores_the_factored_form               PASSED
test_decode_reads_the_block_major_paged_cache[relu]            PASSED
test_decode_reads_the_block_major_paged_cache[factored-fused]  PASSED
test_decode_reads_the_block_major_paged_cache[factored-gather] PASSED
6 passed in 5.34s
```

Against the pre-fix decode kernel, the same inputs give (per row, vs the factored reference so the ReLU is not the variable):

```
row 0 seq_len=  1  token0_ok=False  max_abs_err=5.985  ref_absmax=5.985  within_1pct=0/1
row 1 seq_len= 65  token0_ok=False  max_abs_err=774.4  ref_absmax=36.43  within_1pct=0/65
row 2 seq_len=187  token0_ok=False  max_abs_err=nan    ref_absmax=25.99  within_1pct=1/187
```

The `nan` is a misread scale: four FP8 value bytes reinterpreted as FP32 overflow the fp16 gather.

End-to-end, on `DeepSeek-V4-Flash-0731`, TP8, `max_model_len 262144`:

| | needle 4k–64k (9 pts) | needle 128k/240k (6 pts) | 207k prefill | 128k prefill |
|---|---|---|---|---|
| before | 5/9 EXACT | not measured | 768 s | 268 s |
| after | **9/9 EXACT** | **6/6 EXACT** | **350 s** | **148 s** |

The prefill timings are a single run each; local run-to-run noise on this box is around ±11%, so treat the magnitude as approximate — the direction is well outside it.

Also ran the rest of the SM70 DeepSeek-V4 kernel tests. **20 failures in `tests/kernels/test_fused_indexer_q_rope_quant.py` are pre-existing on `main` @ `7aede2cf01`** — verified by running that file on a clean checkout of `main` in the same environment (20 failed there too). Nothing in this PR changes them.

Lint: `ruff 0.14.0` `check` and `format` clean.

## Notes

- **Not a duplicate.** `gh pr list --repo 1CatAI/1Cat-vLLM --state all --search "indexer relu"` / `"sparse indexer"` returns only #184 (CLOSED) and #159 (MERGED), both 8k bring-up / long-context work that predates this path. No open PR touches `vllm/models/deepseek_v4/sm70/indexer.py`. Both defects are live on `main` today.
- **AI assistance was used** (Claude Opus 5) for the investigation, the patch and the tests. Every changed line has been reviewed by the submitter, and the commands and results above were run on the hardware described.

### DIFF
--- a/tests/kernels/test_deepseek_v4_sm70_indexer.py
+++ b/tests/kernels/test_deepseek_v4_sm70_indexer.py
@@ -4,33 +4,189 @@
 import pytest
 import torch
 
-from vllm.models.deepseek_v4.sm70.indexer import sm70_indexer_prefill_logits
+from vllm.models.deepseek_v4.sm70 import indexer as sm70_indexer
+from vllm.models.deepseek_v4.sm70.indexer import (
+    sm70_indexer_decode_logits,
+    sm70_indexer_prefill_logits,
+)
 
-
-@pytest.mark.skipif(
+requires_sm70 = pytest.mark.skipif(
     not torch.cuda.is_available() or torch.cuda.get_device_capability() != (7, 0),
     reason="requires NVIDIA V100/SM70",
 )
-def test_sm70_indexer_prefill_dequantizes_fp8_storage_as_uint8() -> None:
-    torch.manual_seed(20260803)
-    num_queries = 3
-    num_heads = 4
-    num_keys = 7
-    head_dim = 128
 
-    q = torch.randn(
-        (num_queries, num_heads, head_dim), device="cuda", dtype=torch.float16
+_HEAD_DIM = 128
+_BLOCK_SIZE = 64
+
+# E4M3FN bit patterns with exact values. Building keys from a palette keeps the
+# reference bit-exact and needs no native FP8 conversion, which SM70 lacks.
+_FP8_BITS = torch.tensor([0x00, 0x30, 0x38, 0x3C, 0x40, 0xB0, 0xB8, 0xC0])
+_FP8_VALUES = torch.tensor([0.0, 0.5, 1.0, 1.5, 2.0, -0.5, -1.0, -2.0])
+
+
+def _random_fp8_keys(num_keys: int, generator: torch.Generator):
+    """Return (uint8 bit patterns, exact fp32 values) for `num_keys` keys."""
+    picks = torch.randint(len(_FP8_BITS), (num_keys, _HEAD_DIM), generator=generator)
+    return _FP8_BITS[picks].to(torch.uint8).cuda(), _FP8_VALUES[picks].cuda()
+
+
+def _reference_index_logits(
+    q: torch.Tensor, weights: torch.Tensor, k: torch.Tensor
+) -> torch.Tensor:
+    """I[t, s] = sum_h weights[t, h] * relu(q[t, h] . k[s]).
+
+    The model's actual indexer scoring function (reference `Indexer.forward`,
+    inference/model.py:427; the in-repo `fp8_mqa_logits` reference at
+    v1/attention/ops/rocm_aiter_mla_sparse.py:510 applies the same relu). It
+    does NOT factor to (sum_h weights[t, h] * q[t, h]) . k[s] - the relu sits
+    between the weighting and the head sum.
+    """
+    per_head = torch.einsum("thd,sd->ths", q.float(), k.float())
+    return torch.einsum("ths,th->ts", torch.relu(per_head), weights.float())
+
+
+def _reference_factored_logits(
+    q: torch.Tensor, weights: torch.Tensor, k: torch.Tensor
+) -> torch.Tensor:
+    """The relu-free form, kept reachable via VLLM_SM70_INDEXER_RELU=0."""
+    weighted_q = torch.einsum("thd,th->td", q.float(), weights.float())
+    return weighted_q @ k.float().T
+
+
+def _build_paged_index_cache(
+    value_bits: torch.Tensor, scales: torch.Tensor
+) -> torch.Tensor:
+    """Pack keys into the block-major layout the C++ cache kernels write.
+
+    `indexer_k_quant_and_cache` / `cp_gather_indexer_k_quant_cache`
+    (csrc/libtorch_stable/cache_kernels.cu) store, per block, all `block_size`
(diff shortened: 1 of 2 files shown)


yangzhuxinyzx (approved):
Source audit passed. The corrected byte offsets match both indexer_k_quant_and_cache and cp_gather_indexer_k_quant_cache; per-head ReLU placement matches the official DeepSeek-V4 scoring function. Reviewed prefill tiling, dynamic decode splits, 2-D speculative flattening/block-table replication, and downstream seq-len masking. PR tests pass 6/6 on V100; an additional 64-head multi-tile prefill plus 2-D spec-decode paged-cache probe also passes. diff-check, Ruff check/format, and py_compile pass. No end-to-end rerun per audit scope.

PR #276, opened 2026-08-24
[Bugfix] Zero reused KV blocks in MRV2

wfhe:
## Purpose

Fix stale/non-finite cache data on block reuse in the new MRV2 runner by restoring the existing scheduler-to-runner cache-zero lifecycle contract. This replaces the old 59-file umbrella branch with one focused change rebased on current `main`.

## Source audit

- The reported stale full-attention KV block root cause is valid.
- MRV2 now initializes the shared `KVBlockZeroer` and consumes `SchedulerOutput.new_block_ids_to_zero` before attention or a zero-token early return.
- The path runs only when the scheduler allocates/reuses blocks; it adds no model/checkpoint/architecture selector, per-step sequence-length D2H readback, request-by-layer loop, or token clamp.
- Rejected from the original branch: default-on sampled-token clipping, per-step tail walking, diagnostic-only probes, GDN PAD replication, and last-column gather clamping. Those either mask corruption, add hot-path synchronization/launches, or conceal a broken block-table contract.
- The fix relies only on allocation lifecycle, cache-group/operator layout, block IDs, tensor dtype/shape, and graph-safe zeroing.

## Test Plan

```bash
CUDA_VISIBLE_DEVICES="" UV_PROJECT_ENVIRONMENT=/home/ymzx/miniconda3/envs/vllm-0.0.5-t210 uv run --no-sync pytest -q tests/v1/worker/test_gpu_model_runner_v2_kv_zero.py tests/v1/worker/test_kv_block_zeroer.py
CUDA_VISIBLE_DEVICES="" UV_PROJECT_ENVIRONMENT=/home/ymzx/miniconda3/envs/vllm-0.0.5-t210 uv run --no-sync pytest -q tests/v1/attention/test_gdn_metadata_builder.py -k "align or state_contract"
CUDA_VISIBLE_DEVICES="" UV_PROJECT_ENVIRONMENT=/home/ymzx/miniconda3/envs/vllm-0.0.5-t210 uv run --no-sync pytest -q tests/v1/core/test_prefix_caching.py -k mamba_align
UV_PROJECT_ENVIRONMENT=/home/ymzx/miniconda3/envs/vllm-0.0.5-t210 uv run --no-sync pre-commit run --files vllm/v1/worker/gpu/model_runner.py tests/v1/worker/test_gpu_model_runner_v2_kv_zero.py docs/design/sm70_v100_migration_control.md
```

## Test Result

- MRV2 wiring plus CPU zeroer contracts: 7 passed, 2 CUDA tests skipped under isolated CPU run.
- GDN align/state contracts: 5 passed.
- Mamba align allocator contracts: 2 passed.
- All changed-file pre-commit hooks passed, including ruff, mypy, markdownlint, SPDX, and forbidden-import/API checks.
- The shared CUDA zeroer and its multi-pool interleaved layout were already validated on V100 while auditing merged PR #280; this PR does not change that kernel. No full-model E2E was run.

Signed-off-by: yangzhuxinyzx <153831768+yangzhuxinyzx@users.noreply.github.com>

### DIFF
--- a/docs/design/sm70_v100_migration_control.md
+++ b/docs/design/sm70_v100_migration_control.md
@@ -42788,3 +42788,31 @@ Interpretation:
   graph, cache, and concurrency contracts. The architecture-name check was
   removed; measured checkpoint details remain evidence only and never select
   an optimization route.
+
+## 2026-08-24 PR #276 cache-block reuse audit
+
+- The reported corruption root cause is valid: a reused full-attention cache
+  block can retain non-finite values in slots that its new owner has not yet
+  written, and arithmetic masking does not make `0 * NaN` finite. The MRV2
+  runner had omitted the cache-zero lifecycle hook already used by the legacy
+  runner.
+- MRV2 now initializes the shared `KVBlockZeroer` metadata and consumes the
+  scheduler's `new_block_ids_to_zero` before either attention or a zero-token
+  early return. Work is proportional to newly allocated blocks, uses the
+  cache-group/operator layout metadata, and introduces no per-step sequence
+  length readback or request-by-layer launch loop.
+- The PR's per-step tail walker and sampled-token vocabulary clamp are
+  rejected. The former performs device-to-host synchronization plus many
+  small zero launches on every step; the latter turns corrupted output into
+  an arbitrary boundary token and hides the producer. The proposed GDN PAD
+  replication and last-column gather clamp are also rejected: the align-mode
+  allocator already guarantees a live current block plus its speculative
+  scratch window, so those fallbacks would conceal a broken block-table
+  contract and repeat an unrelated state block.
+- Route selection and correctness depend only on allocation lifecycle,
+  cache-group layout, block IDs, tensor shape/dtype, and operator contracts.
+  No model, checkpoint, or architecture identity participates in the fix.
+- Focused validation covers MRV2 zeroer initialization and zero-before-return
+  ordering, dense and multi-pool interleaved zeroing, Mamba align allocation,
+  and the existing GDN state-window contracts. No full-model end-to-end run
+  was required for this source-level repair.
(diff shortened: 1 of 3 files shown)


yangzhuxinyzx (changes requested):
Request changes on exact head `1ac46ec5e4398c4b9653a9440b0c710def0ad1b1`.

The narrow OOB sentinel kernel itself passes all 9 GPU unit tests on an idle V100, and 17 focused CPU policy/metadata tests pass. However, the integration is not scoped to the PR title and introduces broad default-on correctness/performance behavior:

1. `vllm/v1/worker/gpu/model_runner.py:493-515` builds `_fa_groups` for every non-Mamba 5-D KV group with 1:1 block sizing, without an SM70, DFlash2, hybrid-model, or affected-attention-backend gate. `prepare_attn` then calls `_zero_stale_kv_tail` unconditionally on every step. That function performs two GPU-to-CPU synchronizations, constructs CPU/GPU index tensors, loops every request and layer, and zeroes KV slices. This changes and penalizes unrelated models/platforms. Gate it on the exact affected runtime contract or move the fix to block allocation/reuse where stale storage is created.

2. `model_runner.py:1037` indexes `nb_gpu - 1` before the `nb <= 0` guard at lines 1038-1041. Zero-block requests therefore select column -1 before being filtered, and an empty block-table width can fail outright. Filter valid rows before indexing.

3. `VLLM_SM70_DFLASH2_SAMPLER_SENTINEL` defaults to 1, but `model_runner.py:1859-1863` gates only on `self.speculator is not None`. It therefore runs for Eagle, MTP, DFlash1, other GPUs, and other backends. `input_batch.py:570-580` launches a Triton kernel and immediately calls `event[0].item()`, adding a D2H synchronization to every speculative sampling step, then clamps token IDs. Scope this to exact SM70 DFlash2 or make it default-off.

4. Exact-head static checks fail: Ruff 0.14 reports 6 rule violations and 4 files need formatting. The PR body is still the blank template, so the required Purpose/Test Plan/Test Result and the broad-route validation are missing.

The stacked PR also currently conflicts with its parent branch in `vllm/v1/attention/backends/flash_attn_v100.py`. Resolve the stack, narrow both default-on paths, add negative-route tests, and rerun the exact head before requesting another review.

Issue #433, opened 2026-08-31
[Build][SM70] main (1.5.0-RC state) does not compile for SM70: bf16/fabric code without arch guards in 3 csrc files


SabaTech-dev:
## Environment

- Building main @ fa344edba (1.5.0-RC state, pinned 2026-08-31) from source
- CUDA 12.8 (/usr/local/cuda-12.8), torch 2.10.0+cu128, Python 3.12, driver 570.x
- `TORCH_CUDA_ARCH_LIST=7.0`, `pip install -e . --no-build-isolation`
- Host: 2x Tesla V100-SXM2-32GB

### 1. csrc/quantization/activation_kernels.cu — 41 errors

- `error: identifier "make_bfloat162" is undefined` (line 126, silu2_v2)
- bf16 warp helpers (`__hmax`, `__hmin`, `__hdiv` on `__nv_bfloat16`) and the
  `silu_mul_fp8_quant_deep_gemm_kernel` body use bf16 intrinsics that CUDA only
  exposes for `__CUDA_ARCH__ >= 800`.

### 2. csrc/cumem_allocator.cpp

- `CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED` / `CU_MEM_HANDLE_TYPE_FABRIC`
  undeclared: the TU resolves `<cuda.h>` to the distro toolkit copy
  (`/usr/include/cuda.h`, older) which predates fabric handles.
  Proposed fix (tested locally, works): symbol-guard the fabric block with
  `#ifdef CU_MEM_HANDLE_TYPE_FABRIC` and hoist the `fab_flag` declaration.

### 3. csrc/libtorch_stable/cuda_vec_utils.cuh

- `__float22bfloat162_rn` / `__bfloat1622float2` undefined (lines 334/345) in the
  `cast_to_float2` / `cast_to_packed` `if constexpr` branches, used by
  `_C_stable_libtorch` (built unconditionally for CUDA target).

## Impact

`pip install -e .` for SM70 fails on the 1.5.0-RC state, so the RC cannot be
validated on real V100 hardware before release.

## Suggested direction

The project already has a pattern for this: the ROCm branches in
activation_kernels.cu fall back to `__float22bfloat162_rn(make_float2(...))`-style
emulation. Extending the same arch-conditional emulation (or `#if __CUDA_ARCH__
>= 800` guards around bf16-only kernels that are runtime-dead on Volta anyway,
[...]

PR #435, opened 2026-08-31
build: restore SM70 (Volta) compilation — 3 csrc fixes (fabric symbols, bf16 helpers, arch-conditional activation kernels)

SabaTech-dev:
Restores the SM70 (Volta) build of main (1.5.0-RC state, fa344edba). Evidence and error details in issue #433.

## Changes

1. **csrc/cumem_allocator.cpp** — symbol-guard the fabric-handle block with
   `#ifdef CU_MEM_HANDLE_TYPE_FABRIC` (the TU can resolve `<cuda.h>` to the
   distro toolkit copy, which predates fabric handles) and hoist the
   `fab_flag` declaration so the error handler compiles either way.

2. **CMakeLists.txt** — exclude `csrc/quantization/activation_kernels.cu`
   when none of the requested CUDA architectures is >= 80: its deep_gemm
   fp8 paths instantiate bf16 packed intrinsics that CUDA gates to
   `__CUDA_ARCH__ >= 800` and are runtime-dead on Volta anyway.

3. **csrc/libtorch_stable/cuda_vec_utils.cuh** — add bit-exact SM70
   emulations for the bf16 packed conversions (`vec_bf162_to_float2`,
   `vec_float2_to_bf162_rn`) and route `cast_to_float2` / `cast_to_packed` /
   `packed_mul` through them under `__CUDA_ARCH__ < 800`. bf16 -> fp32 is an
   exact mantissa shift; fp32 mul of two bf16 values is exact and the RNE
   pack yields the correctly-rounded bf16 product (same results as the
   native path). Architectures >= 800 keep the intrinsics untouched.

## Tests run

- `pip install -e . --no-build-isolation` on 2x V100 32GB (SM70-only build,
  CUDA 12.8, torch 2.10.0+cu128, Python 3.12): before = metadata/build
  failure at the first nvcc invocation; after = compiles through the full
  extension set (validation run in progress on real hardware; will report
  back with runtime confirmation on the 1.5.0-RC state).
- `python -m py_compile` on the touched Python-free files is N/A; the
  CMake change was exercised through a clean reconfigure.

## AI assistance disclosure

Prepared with AI coding assistance (OpenCode agent, SabaTech.dev lab). The
submitting human reviewed every changed line and reproduced the SM70 build
failure and the fix on real hardware.

### DIFF
--- a/vllm/platforms/cuda.py
+++ b/vllm/platforms/cuda.py
@@ -486,6 +486,13 @@ class CudaPlatformBase(Platform):
 
     @classmethod
     def use_custom_allreduce(cls) -> bool:
+        # The custom all-reduce kernel is not supported on SM70 (Volta): with
+        # TP > 1 on V100 (PCIe or NVLink) it leads to non-converging piecewise
+        # CUDA-graph capture, decode throughput collapse, and Xid 62 faults on
+        # both GPUs. Returning False makes the parallel config fall back to
+        # NCCL, which is stable on this architecture.
+        if cls.has_device_capability(70) and not cls.has_device_capability(75):
+            return False
         return True
 
     @classmethod


SabaTech-dev:
Status update on this PR: after the recent upstream syncs, the three csrc fixes (fabric symbols, bf16 helpers, arch-conditional activation kernels) were absorbed into main, so the only remaining unique diff here was the SM70 all-reduce patch — the same change as PR #431, which is now the actively maintained one.

PR #431 has been rebased onto current main and re-scoped: it auto-disables custom all-reduce only for SM70 with TP=2 (fixing the startup deadlock from #652) so the SM70-native TP=4/TP=8 push all-reduce routes from #605 keep working.

Unless the maintainers prefer to keep a separate PR for any csrc delta we may have missed, we'd suggest closing this one in favor of #431 — happy to rebase it down to just the absorbed-csrc delta instead if anything is still missing from main. We'll leave the decision to you.

Issue #490, opened 2026-09-04
[Bug][SM70][TP2] 2xV100: decode collapses to ~0.15 tok/s at batch>=2 once per-sequence context exceeds ~221K (Qwen3.8-27B NVFP4, 1.5.0)


areslp:
## Summary

On 2x Tesla V100S-PCIE-32GB (TP2), serving `Qwen3.8-27B-QUASAR-NVFP4` with 1Cat-vLLM 1.5.0,
**decode throughput collapses from ~11-22 tok/s to ~0.15 tok/s as soon as two requests run
concurrently, but only once each sequence's context passes a threshold between 221,184 and
237,568 tokens.**
Batch 1 at the same depth is fine. Prefill is fine. The engine reports
`Running: 2 reqs, Waiting: 0 reqs`, `Avg prompt throughput: 0.0 tokens/s`,
`Avg generation throughput: 0.3 tokens/s` for 14+ minutes straight while both GPUs sit at
100% utilization and ~250 W.

It is deterministic: three independent server launches produced byte-identical wall times
and prefix-cache hit rates.

## Environment

- 2x Tesla V100S-PCIE-32GB (SM70), PCIe, no NVLink, `NCCL_P2P_DISABLE=1`
- driver 580.105.08, CUDA 12.8, torch 2.10.0+cu128, triton 3.6.0, Python 3.12.3
- 1cat-vllm 1.5.0 (`1cat_vllm-1.5.0-cp312-cp312-linux_x86_64.whl`), Ubuntu 24.04 (glibc 2.39)
- Model: `Qwen3.8-27B-QUASAR-NVFP4` (compressed-tensors; log says
  `SM70 compressed-tensors NVFP4 TurboMind W4A16 dense path enabled`, `dtype=torch.float16`)
- `VLLM_SM70_FLASH_ATTN_V100=1`, attention backend `FLASH_ATTN_V100`,
  `SM70 FlashQLA GDN decode route enabled`, `Prefix caching is enabled with Mamba cache 'align' mode`

## Serve command

```bash
vllm serve /path/to/Qwen3.8-27B-QUASAR-NVFP4 --served-model-name qwen38-27b \
  --tensor-parallel-size 2 --max-model-len 262144 --max-num-seqs 3 \
  --kv-cache-dtype fp8_e5m2 --gpu-memory-utilization 0.92 \
  --enable-prefix-caching --host 127.0.0.1 --port 8000
```

Reported at startup:

```
[...]

PR #616, opened 2026-09-13
[Perf][SM70] Pace chunked prefill and unpin the D256 BM32 page size (#490)

areslp:
Rebased onto `main` and narrowed.

The kernel half of the original PR — running the small-query rows of a mixed prefill+decode batch on the paged decode kernels instead of the per-sequence paged prefill kernel — has since landed on `main` in `8a10215d6`, where the flag is on by default. That part is dropped here. What remains is the two pieces that are not in `main`, both default-off:

**1. `VLLM_1CAT_PREFILL_PACE_STEPS=N` (default 0)** — while any running request is decoding, a request still in prompt prefill is given a chunk only every N engine steps, through the existing `Request.next_decode_eligible_step` gate; the other N-1 steps are decode-only. A prefill with nothing else decoding keeps its full budget. This is the GPU-time-sharing half of #490, which the kernel fix does not address: the resident decoder's own row is now cheap, but it still shares each step with the prefill chunk.

**2. `VLLM_FLASH_V100_PREFILL_D256_BM32_ANY_PAGE=1` (default off)** — with MTP the Mamba page gains `num_spec` conv slots, the align-mode attention block becomes 816 instead of 784, and every paged-prefill chunk call fell off the BM32 phase kernel's page-784-only fast path (104 ms instead of 44 ms per 784-token chunk at 104K context). The kernel only used 784 to map each 16-token page slot, so the gate now accepts any multiple of 16 when the flag is set. Page 784 keeps its existing gate and code path.

### Measurements

Pacing, on 2x V100-PCIE-32GB TP2, Qwen3.8-27B-QUASAR-NVFP4, fp16 KV, MTP4, a decoder resident at 240,000 tokens, fixed prompts, zero preemptions:

| pacing N | resident decode during a 16K prefill | 16K prefill wait |
|---|---|---|
| 0 | 4.2 tok/s | 21.3 s |
| 4 | 11.1 tok/s | 31.9 s |
| 8 | 14.6 tok/s | 45.6 s |

@Leonccaa measured the same trade-off independently on a different config (4x V100-PCIE-32GB TP4, Qwen3.8 Flash-Next AWQ, MTP3, FP16 KV, two residents at ~170K/175K, a cold 14,023-token prefill injected):

| arm | resident decode while the 14K prefill runs | 14K TTFT |
|---|---|---|
| all off | 12.7 tok/s | 20.8 s |
| + pacing 4 | 30.0 tok/s | 23.0 s |

That is 2.35x for the resident decoder at +10.8% TTFT for the injected prefill, on a model and quantization this branch was never tuned against.

Any-page: page 816/896 go from 104.4 ms to 43.99 ms standalone, bit-identical to the page-784 path and to a dense reference. End to end (MTP4, resident decoder at 240,000 tokens): a partner 131K prefill drops from 314.7 s to 207.9 s; with pacing 4 it drops from 400.1 s to 293.8 s while the resident decoder runs 9.86 tok/s, up from 7.38.

### Tests

- `tests/v1/core/test_prefill_pacing.py` (new): chunk cadence at N = 0/2/4, the decoder keeping its slot in every step, and a lone prefill staying unpaced.
- `tests/kernels/attention/test_sm70_flash_v100_paged_prefill_any_page.py`: pages 784/816/896 against the dense reference, with the flag off and on.

With both flags unset the scheduler and the kernel dispatch are unchanged.


areslp:
Thanks — that's the arm that matters, and it confirms the split. The rows fix landed on main in `8a10215d6` (now default on), so I've rebased this PR down to the two pieces main doesn't have: the pacing knob and the BM32 any-page gate. Your TP4/AWQ/MTP3 numbers are now quoted in the description as independent validation of the pacing trade-off.

PR #653, opened 2026-09-18
[Bugfix] Constrain tools + response_format with a union grammar in Qwen3XMLToolParser

areslp:
## Purpose

Fixes silently dropped tool calls when `tools` and `response_format` are combined. Same defect as the still-open upstream issue [vllm-project/vllm#39929](https://github.com/vllm-project/vllm/issues/39929).

### The bug

A chat request carrying **both** `tools` and a `response_format` (`json_schema` or `json_object`) with `tool_choice` absent or `"auto"` silently loses its tool calls: `tool_calls` comes back empty, HTTP 200, no warning, and the model returns a **fabricated** schema-conforming answer instead of calling the tool. Output tokens collapse from thousands to a few dozen.

### Root cause

1. `to_sampling_params` (`entrypoints/openai/chat_completion/protocol.py`) converts `response_format` into a grammar constraining the **entire** output, never consulting `tools` / `tool_choice`.
2. `get_json_schema_from_tools()` (`tool_parsers/utils.py`) returns `None` for `"auto"`, so **no tool grammar is built** — `"auto"` tool calling normally works by free generation plus post-hoc text parsing.
3. Only the answer-schema grammar is installed, and the literal `<tool_call>` prefix falls outside its allowed token set. The model is not "choosing not to call" — it **physically cannot emit** the opening token.
4. The asymmetry: `ToolParser.adjust_request` sets the tool schema **and clears `response_format`** on the `required` / named paths, so the tool grammar wins there. Only `"auto"` leaves `response_format` governing decoding.

The `json_object` case shows this most clearly. The model *tries* to call the tool, and the grammar deforms the call into a JSON object returned as plain content:

```json
{"name": "get_secret_token", "arguments": {"city": "Paris"}}
```

The intent survives; the wire format does not, so nothing parses it.

`Qwen3XMLToolParser` overrides neither `adjust_request` nor `get_structural_tag`, so it inherits this wholesale. That is the entire gap — `Qwen3CoderToolParser` already overrides `get_structural_tag` for the same wire format.

### The fix

Override `adjust_request` in `Qwen3XMLToolParser`. It intervenes **only** when the request is a `ChatCompletionRequest` with non-empty `tools`, `tool_choice == "auto"`, and a `json_schema`/`json_object` `response_format`. Everything else returns `super().adjust_request(request)` unchanged.

In that one case it installs an xgrammar `OrFormat` union as `structured_outputs.structural_tag`:

```
output        ::= optional-whitespace ( tool-branch | answer-branch )
tool-branch   ::= ( "<tool_call>\n<function=NAME>\n" params-xml "\n</function>\n</tool_call>" )+   sep "\n"
answer-branch ::= JSON matching response_format
```

and clears `response_format` — mandatory, otherwise `to_sampling_params` derives a second constraint and the single-constraint validation raises through `replace()`.

Both branches stay available on every turn, so a multi-turn agent loop works: tool-call turns early, schema answer on the final turn once tool results are back.

### Three non-obvious design points

**Tool branch shape.** It reuses `get_model_structural_tag(model="qwen_3_5", ...)` with `tool_choice="required"` — a `TagsWithSeparatorFormat` — rather than the `TriggeredTagsFormat` that `"auto"` would produce. Measured on a Qwen3 tokenizer (vocab 248077), allowed tokens at position 0:

| grammar | allowed @ pos 0 | % vocab |
|---|---|---|
| answer schema alone | 2 | 0.00% |
| tools alone, `TagsWithSeparator` | 3 | 0.00% |
| tools alone, `TriggeredTags` | 248075 | 100.00% |
| **union via `TagsWithSeparator`** | **5** | **0.00%** |
| union via `TriggeredTags` | 248075 | 100.00% |

`TriggeredTagsFormat` permits free text before the trigger, so the union's mask degenerates to nearly everything and the answer branch becomes unreachable. This also explains why `"auto"` tool calling works at all today: it is fully unconstrained plus post-hoc parsing.

**Leading whitespace must be allowed.** The chat template emits `</think>\n\n` before content, and with `enable_in_reasoning=False` the grammar binds at the token immediately after `</think>`. Without the whitespace prefix, the tool-vs-answer decision is taken at a token whose natural probability mass is masked out, decided by residual mass instead — that passes smoke tests and picks the wrong branch intermittently under load. With it, position 0 goes from 5 to 164 allowed tokens, still tight, and whitespace alone cannot terminate the match.

**Parallel-call separator.** `structural_tag_registry.py` hardcodes `separator=""` for the `qwen_3_5` `required` branch, but the chat template emits `\n<tool_call>` between parallel calls, so a real `</tool_call>\n<tool_call>` sequence would be rejected and parallel calling would silently collapse to a single call. This PR corrects it at the use site via `model_copy`. The registry itself deviates from both the chat template and xgrammar's own builder here (the `deepseek_v4` builder correctly uses `"\n"`), which also affects the strict-mode `required` path — left for a separate change to keep this PR focused.

### Self-validation and graceful degradation

The union compiles every tool's `parameters` through the `qwen_xml` converter, which the `"auto"` path never did before — a new exposure surface. A schema feature xgrammar cannot express (regex lookahead, for example) would raise and, under `backend="auto"`, fall back to the guidance backend, which raises `KeyError: 'triggers'` on v2 structural tags.

So `adjust_request` self-validates with `Grammar.from_structural_tag` and, on failure, retries with tool `parameters` left unconstrained before finally falling back to the previous behaviour with a warning. The relaxed retry is faithful rather than lossy: today's `"auto"` path does not constrain tool arguments at all, and the XML call structure stays constrained either way.

### Known limitations

- Requests that set the native `structured_outputs` field **without** `response_format` hit the same mechanism and are not fixed here. When `structured_outputs` already carries a constraint, this code deliberately skips the union and defers to the base path rather than stacking a second constraint.
- The Responses API (`text.format`) has the identical bug and is untouched.
- `get_structural_tag` is deliberately **not** added. Under `VLLM_ENFORCE_STRICT_TOOL_CALLING=1` it would activate base Step 1, which installs a structural tag but never clears `response_format`, breaking `required` / named + `response_format` where they work today. `Qwen3CoderToolParser` has that same latent issue; base Step 1 should be fixed before parity is worth restoring.

## Test Plan

Serve a Qwen3 XML tool-calling model with `--enable-auto-tool-choice --tool-call-parser qwen3_xml --reasoning-parser qwen3`, then issue the request below. The tool returns a value the model cannot possibly know, so **any** content-only answer is necessarily fabricated — that makes the failure unambiguous:

```bash
curl -s localhost:8000/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model": "<qwen3-xml-model>",
  "messages": [{"role": "user", "content": "What is the secret token for Paris? Use the tool."}],
  "tools": [{"type": "function", "function": {
    "name": "get_secret_token",
    "description": "Returns the random secret token for a city.",
    "parameters": {"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}}}],
  "response_format": {"type": "json_schema", "json_schema": {"name": "a", "strict": true, "schema": {
    "type": "object", "properties": {"token": {"type": "string"}},
    "required": ["token"], "additionalProperties": false}}},
  "temperature": 0
}'
```

The full matrix sweeps the same request across `response_format` in {none, `json_schema` strict, `json_schema` lax, `json_object`, `{"type":"text"}`} × `tool_choice` in {absent, `"auto"`, `"required"`, named, `"none"`, explicit `null`}, plus streaming, two-tool, no-tools, and a no-tool-needed prompt that must still produce a schema answer.

## Test Result

15-case matrix against a live server (`temperature: 0`, `seed: 42`):

| # | `response_format` | `tool_choice` | before | after |
|---|---|---|---|---|
| 01 | none | absent | 1 call | 1 call |
| 02 | json_schema strict | absent | **0, fabricated** | **1 call** |
| 03 | json_schema lax | absent | **0, fabricated** | **1 call** |
| 04 | json_object | absent | **0, call deformed into content** | **1 call** |
| 05 | `{"type":"text"}` | absent | 1 call | 1 call |
| 06 | json_schema strict | `"auto"` | **0, fabricated** | **1 call** |
| 07 | json_schema strict | `"required"` | 1 call | 1 call |
| 08 | json_schema strict | named | 1 call | 1 call |
| 09 | json_schema strict | absent, streaming | **0 deltas** | **1 logical call** |
| 10 | json_schema strict | `"none"` | schema answer | unchanged |
| 11 | none | `"none"` | raw XML in content | unchanged |
| 12 | **answer-branch** schema | absent | schema answer | **schema answer** |
| 13 | json_schema strict, 2 tools | absent | **0, fabricated token** | **1 call** |
| 14 | no tools + schema | — | schema answer | unchanged |
| 15 | json_schema strict | explicit `null` | schema answer | unchanged |

7 broken cases fixed, 8 working cases unchanged. Case 12 is the important one: the answer branch stays reachable when no tool is needed, so the union genuinely selects a branch rather than always taking the tool path.

Additional live checks:

- Parallel calls with `response_format` return 2 calls (`Paris` + `London`) — this is what the separator correction buys.
- A tool whose parameter carries a regex lookahead still returns a correct call via the relaxed retry, with the warning logged; no hard fallback, no traceback.
- No tracebacks, `KeyError`, or `Invalid structural tag` in the server log across the whole sweep.

Offline, 31 assertions against the real tokenizer cover branch acceptance and rejection (`{"nope": 1}` rejected), stop-token reachability after each branch, the mask counts in the table above, the `enable_in_reasoning=True` wrapper, and **byte-identical** equivalence with the base `adjust_request` for every non-intervened request shape (auto without `response_format`, `required`, named, `"none"`).


areslp:
`pre-run-check` is failing only because this is my first PR here (0 merged, and I don't have triage rights to add a label myself), so `pre-commit` has not run yet.

To save a round-trip, I ran the same hooks locally at the stage CI uses (`pre-commit run --hook-stage manual`) against the changed file: ruff check, ruff format, typos, mypy on Python 3.10 through 3.13, SPDX headers, root lazy imports, forbidden imports and the rest all pass. The commit also carries a DCO `Signed-off-by`.

Could a maintainer add the `ready` (or `verified`) label so CI can run? Happy to address anything CI or review turns up.

PR #684, opened 2026-09-23
[Perf] Reduce Qwen3.8 MTP4 verification handoff latency

yangzhuxinyzx:
## Purpose

Reduce the complete Qwen3.8 NVFP4 MTP4 verification round on TP4 V100 while preserving PLE bytes, model arithmetic, and output tokens. The PR includes short disk-backed PLE mmap gathers, short CPU n-gram ID generation, direct pinned four-rank H2D fanout, earlier PLE submission, the active GDN metadata row-selection fast path, and shared batch-level GDN metadata for MTP4. A further grouped GDN write is implemented behind an opt-in flag pending a verifier-trajectory audit. Larger prefill gathers retain their deduplicated parallel route; no 47.684-GiB anonymous table is allocated.

## Test plan

- Byte-exact PLE tests for 0/1/80/128/129/256 rows, duplicate/boundary/invalid IDs, randomized EOS and padded short-ID batches, and four-rank H2D/semaphore behavior.
- GDN pure-spec, mixed, and pure non-spec CPU/GPU row-selection equality tests; MTP4 shared metadata and fused group metadata at query lengths 2-5 in both cache modes; existing DFlash2 fusion tests; full GDN metadata test file.
- Matched TP4/V2 full-model test on V100 GPUs 0-3: 8K prompt, 513 greedy outputs, MTP4, FP16 activations/KV, disk mmap PLE without prefault, and CUDA Graph. Compare full decode/round latency, 306 closed trace intervals, draft counts, and token IDs for two fixed requests and three natural prompts.
- Compare same-source default shared path with a full-model old-path rollback; keep the fused route opt-in and the PR Draft until its acceptance-trajectory difference is understood.

## Test result

- The three previously affected test files passed 188 tests; the expanded full GDN metadata file passes **98 tests**. Ruff check/format and `git diff --check` pass.
- Under the same low-overhead graph capture, draft-combine to next PLE input fell from **4.951 to 3.248 ms** mean across 306 closed rounds (−34.4%). The complete captured round fell from 40.669 to 36.865 ms mean.
- Untraced fixed-request complete-round mean fell from **37.069 to 35.270 ms** (11.38005 to 10.82794 s / 307 rounds). Both fixed outputs and all three natural-prompt outputs match the original control token-for-token, with identical verifier-round counts. GPU processes exited.
- The fixed-slice/no-copy GDN candidate previously failed whole-model output parity and remains excluded. This PR's GDN path preserves an independent contiguous copy; it passed the full-model token gate.
- On the same current source, old-path and default shared-GDN runs both used **306 fixed-prompt verification rounds**, and all five case outputs matched token-for-token. Matched untraced second fixed-request decode fell from **10.756 to 10.032 s** (47.60 to 51.04 emitted tok/s, +7.2%). `VLLM_SM70_MTP4_SHARED_GDN_METADATA=0` rolls back the shared route.
- The opt-in `VLLM_SM70_MTP4_FUSED_GDN_METADATA=1` route cut the draft-combine to next-PLE trace mean from **3.248 to 0.896 ms** and kept all final token IDs exact. However, it used **308** fixed-request verifier rounds versus the current-source control's 306. It remains **default-off** until that difference is explained; its whole-round metrics are not an exact-trajectory A/B.
- The 20-ms complete-round goal is not met. The final captured post-PLE target and sampling/draft windows alone total 31.292 ms; further optimization must target those costs too. Exact configuration and local trace evidence are documented in `docs/design/sm70_v100_migration_control.md`.


Leonccaa:
We ran an isolated V100 A/B against our current Qwen3.8 Flash Next serving stack to check whether this PR helps a pinned-host PLE configuration.

- Hardware/config: 4 × V100-PCIE-32GB (200 W cap), TP4, AWQ g32, E4M3 KV, MTP3 greedy, `FULL_AND_PIECEWISE` CUDA Graph, `max_num_seqs=4`, pinned-host PLE. Disk PLE and the PLE CPU offload connector are inactive in this configuration.
- Single variable: the same serving image and model in both arms; the candidate read-only mounted `vllm/v1/attention/backends/gdn_attn.py` from PR commit `d925bb29ce`. The baseline file was byte-identical to that commit's pre-change GDN file. All other image files and native modules passed the release hash checks in both arms.
- Method: identical 1,024- and 8,192-token prompts, 384 greedy output tokens, one discarded shape warm-up and seven measured requests per arm. Each request used a distinct prefix-cache salt. Decode throughput is measured from first to last streamed token; startup and prefill are excluded.

| Prompt | Baseline median | GDN candidate median | Change |
| --- | ---: | ---: | ---: |
| 1,024 tokens | 67.622 tok/s | 69.152 tok/s | +2.26% |
| 8,192 tokens | 55.925 tok/s | 57.062 tok/s | +2.03% |

For both prompts, all seven streamed outputs had the same text hash across arms, and a separate full response matched all 384 token IDs exactly. Streaming chunk counts also matched (113 and 156 respectively). Median TTFT was effectively unchanged. Neither arm recorded errors or KV preemptions.

We also ran three C4 batches of four 1,024-token/256-output requests. The baseline aggregate throughput varied widely (83.929–152.282 tok/s), so these batches do not support a reliable concurrency-speedup claim.

This is one model start per arm, with sequential arms and a synthetic fixed prompt. It supports a roughly 2% single-request decode gain from the GDN part on this stack; it does not validate the PR's disk-PLE/MTP4 gains or a production workload mix. We did not deploy the candidate.

The PR advanced to `98b81ea69c` during our run. We did not V100-test that newer commit. Its added shared/fused GDN metadata path is gated on MTP4 (the fused variant is also opt-in), so our MTP3 result is limited to the earlier GDN change and should not be read as evidence for the new path.

